PR #70, opened 2021-11-28
Added auto-install powershell script for Windows users

SpectrixDev:
I added a script that automatically downloads the zipped folder straight from the main branch, unzips it and then applies the changes to spicetify. It also checks if spicetify is installed or not.

Note: This only supports windows, it would probably be a good idea if there were a linux version of this too

Feel free to change anything unnecessary in here. This was edited from https://github.com/JulienMaille/dribbblish-dynamic-theme/blob/main/install.ps1 

Love this project so far, I just thought it'd be nice to add an auto install script 👌

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -1,0 +1,129 @@
+# Adapted from https://github.com/JulienMaille/dribbblish-dynamic-theme/blob/main/install.ps1
+# Edited from project Denoland install script (https://github.com/denoland/deno_install)
+
+param (
+  [string] $version
+)
+
+$PSMinVersion = 3
+
+if ($v) {
+  $version = $v
+}
+
+# Helper functions for pretty terminal output.
+function Write-Part ([string] $Text) {
+  Write-Host $Text -NoNewline
+}
+
+function Write-Emphasized ([string] $Text) {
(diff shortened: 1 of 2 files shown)


install.ps1
```diff
@@ -0,0 +1,131 @@
...
+  Write-Done
+ 
+  # Installing.
+  Write-Part "INSTALLING";
+  cd $sp_dot_dir
+  spicetify config custom_apps spicetify-marketplace
+  spicetify apply
+  Write-Done
+
+  # Add patch
```
theRealPadster, line 101: We should probably remove this patch part. I believe it's added by dribbblish dynamic to disable the auto loading of sidebar items, at least that's what it seems to say in the readme. We don't need that for marketplace.

theRealPadster:
Thanks for this! I don't know anything about PowerShell (only use Windows for gaming), so won't be able to help if anyone needs support or has any issues with the script. We can add it though, if you've tested it and it's all good?

SpectrixDev:
There seems to be issues with this, I got a bunch of my friends to test it. I'm going to work on it and also change things based on your suggestions. I'll update it soon

SpectrixDev:
I've made too many mistakes doing this, I am going to close the pull request and make another one when I'm done lol

PR #88, opened 2021-12-28
Update package lock

theblobscp:
When I install dependcies through name, it says the lock file is old. This pr fixes that

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,9021 @@
 {
+  "name": "spicetify-marketplace",
+  "lockfileVersion": 2,
   "requires": true,
-  "lockfileVersion": 1,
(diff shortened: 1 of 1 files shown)


theRealPadster:
Which node version are you using? It's set up with 14.18.1 currently, which uses `npm` v6.x, which is still on [lockfileVersion](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#version) 1. I'm pretty sure I tried updating to use v16 at some point, and my VSCode gave me a ton of linting/typescript errors until I changed it back to 14.

theRealPadster:
Hmm, now it seems to work. I'll merge and give it a try

PR #92, opened 2021-12-29
Use yarn

theblobscp:
Yarn is in many ways a better package manager. This pr aims to use it instead of npm


theRealPadster:
I do like yarn better tbh, but I figured npm is more accessible for people to develop with since most people are more comfortable with npm. The benefits aren't really worth it.

PR #93, opened 2021-12-29
Fix windows path

theblobscp:
Assuming you installed using the binary, this is the path you should use.


theRealPadster:
`.spicetify` is what's in the [documentation](https://spicetify.app/docs/getting-started/custom-apps). As far as I can tell, spicetify-cli is the app folder, and .spicetify is the user folder, where any user-installed content should go.

PR #117, opened 2022-02-06
Optimization

CharlieS1103:
Fixed the PR!

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -342,8 +342,7 @@ async function appendInformationToLocalStorage(array, type) {
             await sleep(10000);
             let themes = await fetchThemes(repo.contents_url, repo.default_branch);
 
-            // TODO Implement checks to make sure the theme is valid
-            if (themes[0]) {
+            if (themes) {
                 addToSessionStorage(themes);
             }
         }
@@ -351,8 +350,7 @@ async function appendInformationToLocalStorage(array, type) {
         for (const repo of array.items) {
             await sleep(10000);
             let extensions = await fetchExtensions(repo.contents_url, repo.default_branch);
-            // TODO: Implement checks to make sure it's a valid extension
-            if (extensions[0]) {
+            if (extensions) {
                 addToSessionStorage(extensions);
             }
 
@@ -371,8 +369,10 @@ async function fetchThemes(contents_url, branch) {
         if (!Array.isArray(manifests)) manifests = [manifests];
         manifests.user = user;
         manifests.repo = repo;
-
-        return manifests;
+        if (manifests[0] && manifests[0].name && manifests[0].usercss && manifests[0].description) {
+            return manifests;
+        }
+        return null;
     }
     catch (err) {
         // console.warn(contents_url, err);
@@ -392,7 +392,10 @@ async function fetchExtensions(contents_url, branch) {
         if (!Array.isArray(manifests)) manifests = [manifests];
         manifests.user = user;
         manifests.repo = repo;
-        return manifests;
+        if (manifests[0] && manifests[0].name && manifests[0].description && manifests[0].main) {
+            return manifests;
+        }
+        return null;
     }
     catch (err) {
         // console.warn(contents_url, err);
--- a/extension.js
+++ b/extension.js
@@ -339,19 +339,18 @@ async function appendInformationToLocalStorage(array, type) {
     // This system should make it so themes and extensions are stored concurrently
     if (type == "theme") {
         for (const repo of array.items) {
-            await sleep(10000);
             let themes = await fetchThemes(repo.contents_url, repo.default_branch);
-
             if (themes) {
                 addToSessionStorage(themes);
+                await sleep(10000);
             }
         }
     } else if (type == "extension") {
         for (const repo of array.items) {
-            await sleep(10000);
             let extensions = await fetchExtensions(repo.contents_url, repo.default_branch);
             if (extensions) {
                 addToSessionStorage(extensions);
+                await sleep(10000);
             }
 
         }
--- a/extension.js
+++ b/extension.js
@@ -456,7 +456,7 @@ async function fetchExtensions(contents_url, branch, stars) {
             };
             if (manifest && manifest.name  && manifest.description) {
                 accum.push(item);
-                console.log("Pushed to accum");
+
             }
             addToSessionStorage(accum);
             return accum;
@@ -474,8 +474,14 @@ async function getRepoManifest(user, repo, branch) {
 }
 // make a function to add an array of items to session storage
 function addToSessionStorage(items) {
+
     items.forEach(item => {
-        window.sessionStorage.setItem(item.title+"-"+item.repo, JSON.stringify(item));
+        const existing = window.sessionStorage.getItem(item.user + item.repo);
+
+        const parsed = existing ? JSON.parse(existing) : [];
+        parsed.push(item);
+        window.sessionStorage.setItem(item.user + "-" + item.repo, JSON.stringify(parsed));
+
     });
 
 }
--- a/extension.js
+++ b/extension.js
@@ -297,38 +297,59 @@ const getParamsFromGithubRaw = (url) => {
 
 const ITEMS_PER_REQUEST = 100;
 
-async function queryThemeRepos() {
-    const BLACKLIST = window.sessionStorage.getItem("marketplace:blacklist");
-    let url = `https://api.github.com/search/repositories?q=${encodeURIComponent("topic:spicetify-themes")}&per_page=${ITEMS_PER_REQUEST}`;
-    const allRepos = await fetch(url).then(res => res.json()).catch(() => []);
-    if (!allRepos.items) {
-        Spicetify.showNotification("Too Many Requests, Cool Down.");
-    }
-    const filteredResults = {
-        ...allRepos,
-        items: allRepos.items.filter(item => !BLACKLIST.includes(item.html_url)),
-    };
-    return filteredResults;
-}
-
-async function queryExtensionRepos() {
-    const BLACKLIST = window.sessionStorage.getItem("marketplace:blacklist");
-    let url = `https://api.github.com/search/repositories?q=${encodeURIComponent("topic:spicetify-extensions")}&per_page=${ITEMS_PER_REQUEST}`;
-    const allRepos = await fetch(url).then(res => res.json()).catch(() => []);
-    if (!allRepos.items) {
-        Spicetify.showNotification("Too Many Requests, Cool Down.");
-    }
-    const filteredResults = {
-        ...allRepos,
-        items: allRepos.items.filter(item => !BLACKLIST.includes(item.html_url)),
-    };
-    return filteredResults;
-}
-
 async function Blacklist() {
     const url = "https://raw.githubusercontent.com/CharlieS1103/spicetify-marketplace/main/blacklist.json";
     const jsonReturned = await fetch(url).then(res => res.json()).catch(() => {});
     return jsonReturned.repos;
+}
+
+/**
+ * TODO
+ * @param {"theme"|"extension"} type The repo type
+ * @param {number} pageNum The page number
+ * @returns TODO
+ */
+async function queryRepos(type, pageNum = 1) {
+    const BLACKLIST = window.sessionStorage.getItem("marketplace:blacklist");
+
+    let url = `https://api.github.com/search/repositories?per_page=${ITEMS_PER_REQUEST}`;
+    if (type === "extension") url += `&q=${encodeURIComponent("topic:spicetify-extensions")}`;
+    else if (type === "theme") url += `&q=${encodeURIComponent("topic:spicetify-themes")}`;
+    if (pageNum) url += `&page=${pageNum}`;
+
+    const allRepos = await fetch(url).then(res => res.json()).catch(() => []);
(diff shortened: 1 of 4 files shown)


extension.js
```diff
@@ -294,3 +294,135 @@ const getParamsFromGithubRaw = (url) => {
     const installedExtensions = getLocalStorageDataFromKey(LOCALSTORAGE_KEYS.installedExtensions, []);
     installedExtensions.forEach((extensionKey) => initializeExtension(extensionKey));
 })();
+const ITEMS_PER_REQUEST = 100;
+async function storeThemes() {
+    const BLACKLIST = await Blacklist();
```
theRealPadster, line 299: In `storeExtensions` and `storeThemes`, it calls the `Blacklist()` twice. We could likely just call it once somewhere else and have it cached.

extension.js
```diff
@@ -294,3 +294,135 @@ const getParamsFromGithubRaw = (url) => {
...
+            if (extensions) {
+                addToSessionStorage(extensions);
+                await sleep(10000);
+            }
+
+        }
+    }
+}
+// This function is used to fetch manifest of a theme and return it
+async function fetchThemes(contents_url, branch) {
```
theRealPadster, line 361: Can we rename this to `fetchThemeManifest` or something? The name is confusing with `storeThemes`.

extension.js
```diff
@@ -294,3 +294,135 @@ const getParamsFromGithubRaw = (url) => {
...
+        }
+        return null;
+    }
+    catch (err) {
+        // console.warn(contents_url, err);
+        return null;
+    }
+}
+// This function is used to fetch manifest of an extension and return it
+async function fetchExtensions(contents_url, branch) {
```
theRealPadster, line 383: Can we rename this to `fetchExtensionManifest` or something? The name is confusing with `storeExtensions`.

extension.js
```diff
@@ -294,3 +294,135 @@ const getParamsFromGithubRaw = (url) => {
...
+    const filteredResults = {
+        ...allRepos,
+        items: allRepos.items.filter(item => !BLACKLIST.includes(item.html_url)),
+    };
+    return filteredResults;
+}
+async function storeExtensions() {
+    const BLACKLIST = await Blacklist();
+    let url = `https://api.github.com/search/repositories?q=${encodeURIComponent("topic:spicetify-extensions")}&per_page=${ITEMS_PER_REQUEST}`;
+    const allRepos = await fetch(url).then(res => res.json()).catch(() => []);
```
theRealPadster, line 314: This only fetches the first 100 repos returned. It should loop somehow until there are no results left. Something like how `loadPage()` works in `index.js` might work. Same goes for `storeThemes()`. 

```js
const currentPage = requestPage > -1 && requestPage ? requestPage : 1;
// Sets the amount of items that have thus been fetched
const soFarResults = ITEMS_PER_REQUEST * (currentPage - 1) + pageOfRepos.page_count;
const remainingResults = pageOfRepos.total_count - soFarResults;

// If still have more results, return next page number to fetch
console.log(`Parsed ${soFarResults}/${pageOfRepos.total_count} extensions`);
if (remainingResults > 0) return currentPage + 1; // There are more results. currentPage + 1 is the next page to fetch. 
else console.log("No more extension results");
```

extension.js
```diff
@@ -294,3 +294,135 @@ const getParamsFromGithubRaw = (url) => {
...
+async function Blacklist() {
+    const url = "https://raw.githubusercontent.com/CharlieS1103/spicetify-marketplace/main/blacklist.json";
+    const jsonReturned = await fetch(url).then(res => res.json()).catch(() => { });
+    return jsonReturned.repos;
+}
+(async function initializePreload() {
+
+    // Begin by getting the themes and extensions from github
+    const extensionsArray = await storeExtensions();
+    const themesArray = await storeThemes();
```
theRealPadster, line 333: Maybe run them concurrently. 
```suggestion
    const [extensionsArray, themesArray] = await Promise.all([
        storeExtensions(), 
        storeThemes(),
    ]);
```

theRealPadster, line 333: These could also be named something like "extensionReposArray", to be more clear.

extension.js
```diff
@@ -294,3 +294,135 @@ const getParamsFromGithubRaw = (url) => {
...
+    } else if (type == "extension") {
+        for (const repo of array.items) {
+            let extensions = await fetchExtensions(repo.contents_url, repo.default_branch);
+            if (extensions) {
+                addToSessionStorage(extensions);
+                await sleep(10000);
+            }
+
+        }
+    }
```
theRealPadster, line 358: Minor cleanup. 
```suggestion
    for (const repo of array.items) {
        const data = (type === "theme")
            ? await fetchThemes(repo.contents_url, repo.default_branch)
            : await fetchExtensions(repo.contents_url, repo.default_branch);
        if (data) {
            addToSessionStorage(data);
            await sleep(10000);
        }
    }
```

extension.js
```diff
@@ -294,3 +294,135 @@ const getParamsFromGithubRaw = (url) => {
...
+        parsed.push(item);
+        window.sessionStorage.setItem(key, JSON.stringify(parsed));
+
+    });
+
+}
+// This function is used to sleep for a certain amount of time
+function sleep(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+}
```
theRealPadster, line 428:
```suggestion
}

```

index.js
```diff
@@ -615,8 +615,15 @@ async function getRepos(page = 1) {
  * @returns The manifest object
  */
 async function getRepoManifest(user, repo, branch) {
-    const url = `https://raw.githubusercontent.com/${user}/${repo}/${branch}/manifest.json`;
+    if (window.sessionStorage.getItem(user + "-" + repo)) {
+        return JSON.parse(window.sessionStorage.getItem(user + "-" + repo));
```
theRealPadster, line 619: This makes it only grab from sessionStorage once. 
```suggestion
    const sessionStorageItem = window.sessionStorage.getItem(`${user}-${repo}`);
    if (sessionStorageItem) {
        return JSON.parse(sessionStorageItem);
```

index.js
```diff
@@ -778,3 +791,7 @@ async function getThemeRepos(page = 1) {
 
     return filteredResults;
 }
+function getUnloadableManifests() {
+    const unloadableManifests =  window.sessionStorage.getItem("noManifests");
+    return  JSON.parse(unloadableManifests);
+}
```
theRealPadster, line 802:
```suggestion
}

```

index.js
```diff
@@ -778,3 +791,7 @@ async function getThemeRepos(page = 1) {
 
     return filteredResults;
 }
+function getUnloadableManifests() {
```
theRealPadster, line 799: This function (and the session storage query from `getRepoManifests`) could be replaced by a single `getSessionStorageValue` function like the one from `index.js`. 

```js
/**
 * Get localStorage data (or fallback value), given a key
 * @param {string} key The localStorage key
 * @param {any} fallback Fallback value if the key is not found
 * @returns The data stored in localStorage, or the fallback value if not found
 */
const getLocalStorageDataFromKey = (key, fallback) => {
    const str = localStorage.getItem(key);
    if (!str) return fallback;

    const obj = JSON.parse(str);
    return obj;
};
```

index.js
```diff
@@ -615,8 +615,15 @@ async function getRepos(page = 1) {
...
+    if (window.sessionStorage.getItem(user + "-" + repo)) {
+        return JSON.parse(window.sessionStorage.getItem(user + "-" + repo));
+    }
 
+    const url = `https://raw.githubusercontent.com/${user}/${repo}/${branch}/manifest.json`;
+    const unloadable = getUnloadableManifests();
+    if ( unloadable.includes(url)) {
+        return null;
+    }
     return await fetch(url).then(res => res.json()).catch(() => null);
```
theRealPadster, line 627: This should probably cache the value to sessionStorage as well, after it's go it, to reduce calls.

index.js
```diff
@@ -698,11 +705,17 @@ async function fetchRepoExtensions(contents_url, branch, stars) {
...
 async function fetchThemes(contents_url, branch, stars) {
     try {
+        let manifests;
         const regex_result = contents_url.match(/https:\/\/api\.github\.com\/repos\/(?<user>.+)\/(?<repo>.+)\/contents/);
         // TODO: err handling?
         if (!regex_result || !regex_result.groups) return null;
         let { user, repo } = regex_result.groups;
-        let manifests = await getRepoManifest(user, repo, branch);
+        if (window.sessionStorage.getItem(user + "-" + repo)) {
+            manifests = JSON.parse(window.sessionStorage.getItem(user + "-" + repo));
```
theRealPadster, line 719: Same optimization comment as `getRepoManifest`

index.js
```diff
@@ -631,11 +638,16 @@ async function getRepoManifest(user, repo, branch) {
...
     try {
         // TODO: use the original search full_name ("theRealPadster/spicetify-hide-podcasts") or something to get the url better?
+        let manifests;
         const regex_result = contents_url.match(/https:\/\/api\.github\.com\/repos\/(?<user>.+)\/(?<repo>.+)\/contents/);
         // TODO: err handling?
         if (!regex_result || !regex_result.groups) return null;
         const { user, repo } = regex_result.groups;
-        let manifests = await getRepoManifest(user, repo, branch);
+        if (window.sessionStorage.getItem(user + "-" + repo)) {
+            manifests = JSON.parse(window.sessionStorage.getItem(user + "-" + repo));
```
theRealPadster, line 647: Same optimization comment as `getRepoManifest`

index.js
```diff
@@ -698,11 +710,17 @@ async function fetchRepoExtensions(contents_url, branch, stars) {
...
+        let manifests;
         const regex_result = contents_url.match(/https:\/\/api\.github\.com\/repos\/(?<user>.+)\/(?<repo>.+)\/contents/);
         // TODO: err handling?
         if (!regex_result || !regex_result.groups) return null;
         let { user, repo } = regex_result.groups;
-        let manifests = await getRepoManifest(user, repo, branch);
+        if (window.sessionStorage.getItem(user + "-" + repo)) {
+            manifests = JSON.parse(window.sessionStorage.getItem(user + "-" + repo));
+        } else {
+            manifests = await getRepoManifest(user, repo, branch);
```
theRealPadster, line 721: This is the same thing the `extension.js` is caching, I think? So we could cache the results here as well?

index.js
```diff
@@ -631,11 +638,16 @@ async function getRepoManifest(user, repo, branch) {
...
+        let manifests;
         const regex_result = contents_url.match(/https:\/\/api\.github\.com\/repos\/(?<user>.+)\/(?<repo>.+)\/contents/);
         // TODO: err handling?
         if (!regex_result || !regex_result.groups) return null;
         const { user, repo } = regex_result.groups;
-        let manifests = await getRepoManifest(user, repo, branch);
+        if (window.sessionStorage.getItem(user + "-" + repo)) {
+            manifests = JSON.parse(window.sessionStorage.getItem(user + "-" + repo));
+        } else {
+            manifests = await getRepoManifest(user, repo, branch);
```
theRealPadster, line 649: This is the same thing the `extension.js` is caching, I think? So we could cache the results here as well?

theRealPadster (changes requested):
Added my notes. Looking pretty good, but added a few tweaks, and there are a couple changes that should probably be made. Thanks for tackling this issue! :)

extension.js
CharlieS1103: @theRealPadster Done (made a call in initalizePreload to add blacklist to session storage)

index.js
```diff
@@ -580,7 +580,7 @@ class Grid extends react.Component {
  * @param {number} page The query page number
  * @returns Array of search results (filtered through the blacklist)
  */
-async function getRepos(page = 1) {
+async function getExtensionRepos(page = 1) {
```
theRealPadster, line 583: Let's call this guy `fetchExtensionRepos` to keep it the same as `fetchThemeRepos`

index.js
```diff
@@ -615,16 +615,20 @@ async function getRepos(page = 1) {
  * @returns The manifest object
  */
 async function getRepoManifest(user, repo, branch) {
-    if (window.sessionStorage.getItem(user + "-" + repo)) {
-        return JSON.parse(window.sessionStorage.getItem(user + "-" + repo));
+    const sessionStorageItem = window.sessionStorage.getItem(`${user}-${repo}`);
+    const failedSessionStorageItems = window.sessionStorage.getItem(`noManifests`);
```
theRealPadster, line 619: It's just a string. 
```suggestion
    const failedSessionStorageItems = window.sessionStorage.getItem("noManifests");
```

index.js
```diff
@@ -796,7 +795,15 @@ async function getThemeRepos(page = 1) {
...
+function addToSessionStorage(items, key) {
+    if (!items || items == null) return;
+    items.forEach(item => {
+        if (!key) key = `${items.user}-${items.repo}`;
+        // If the key already exists, it will append to it instead of overwriting it
+        const existing = window.sessionStorage.getItem(key);
+        const parsed = existing ? JSON.parse(existing) : [];
+        parsed.push(item);
+        window.sessionStorage.setItem(key, JSON.stringify(parsed));
+
```
theRealPadster, line 807: Random empty line. 
```suggestion
```

theRealPadster (approved):
It's probably fine now

Issue #116, opened 2022-02-05
Expose useful methods in the global scope


theRealPadster:
Like setting `window.Marketplace = { reset: ourResetFn };` should allow for running `Marketplace.reset()` in the console if the UI is broken.

PR #118, opened 2022-02-06
Expose Marketplace object in dev console

theRealPadster:
This exposes a global `Marketplace` object. I added a `reset()` function. 
image
Addresses #116

Are there any other things we should add? I was thinking maybe something that generates a list of marketplace settings and outputs it into a standard format, which could help debug things.


theRealPadster:
We could maybe add a function to just reset theme or extensions. Like `Marketplace.reset({ themes: true, extensions: true, tabs: true });`
Not sure if that's worth it though, probably reset all is enough.

CharlieS1103:
@theRealPadster Looks all good to me

PR #126, opened 2022-02-19
Custom css input

CharlieS1103:
Loaded in as a snippet


theRealPadster (approved):
I think we're good, no?

CharlieS1103:
@theRealPadster It has some merge conflicts with the styling I believe.

theRealPadster:
Looks like it's just the map file. Should probably merge the latest updates in

PR #134, opened 2022-02-27
New snippet

ohitstom:


CharlieS1103:
@theRealPadster Does this seem like it fits the use case of snippets? I feel like it's different from the intended use of snippets but still rather cool.

ohitstom:
Would be awesome if you could let this through even if it doesnt fit the use case, its still a css snippet after all, i'd just rather not have to go through the effort of creating a custom theme menu and stuff

CharlieS1103 (approved):
Seems all good,

PR #150, opened 2022-03-27
Add header and grid for each type of card

### LINKED ISSUE #87: Snippest are wrong size in installed tab
We should either A: Create a different way to install and uninstall snippets (maybe a checkbox list but meh) B: Put all the snippets before or after everything in installed tab

theRealPadster:
Add a header and a separate grid for each type of card in the `render()` function. This allows card types to be separated in the Installed tab, which fixes #87


CharlieS1103 (approved):
All looks good to me! Might need to clean up those .map's but I'm really unsure how.

theRealPadster:
> 

The maps are auto generated. They shouldn't be tracked in git tbh. I'll merge now.

PR #151, opened 2022-03-27
Add basis of tags system

### LINKED ISSUE #66: Use better indication for external JS in theme cards
e.g. Add an icon or something. Something that looks better than the label at the bottom of the card.

theRealPadster:
This PR adds a tag generation function and styling. Tags can be specified in the manifest (`tags?: string[]`), and it automatically adds an "external JS" tag if the manifest has the `include` field. 

```json
{
    "name": "My Theme",
    "description": "Sample theme for tags.",
    "tags": [ "minimal", "dark"],
}
```

Currently looks like this:
image

We can potentially add colours or other styling for specific tags if we want, for example:
image

This would serve to resolve #66 

In the future, it would be nice if it was possible to click on tags and filter by them. But that's a whole lot more effort, for another day.


theRealPadster:
Maybe we should add a settings option for hiding tags (other than external JS). Or maybe we can set a max limit for them. Or like a "..." button that will load more. Just in case someone adds like 25 and makes that row of cards look all messed up.

CharlieS1103:
Yeah we can do the overflow thing with css probably, but definitely think we should make it optional

theRealPadster:
Example of how it looks. I made the limit 4. This includes the external JS tag. If there are more than 4, it will show the "..." button below the tags. If I do it inline with the tags, it can get stuck underneath the install/remove button. 

image

index.js
```diff
@@ -100,16 +100,18 @@ try {
 
 // eslint-disable-next-line no-redeclare
 const CONFIG = {
```
CharlieS1103, line 102: Could we potentially just put all of the settings in the same localstorage JSON rather than having a different one for each Setting option?

src/Card.js
```diff
@@ -77,13 +82,18 @@ class Card extends react.Component {
 
         Object.assign(this, props);
 
+        // Needs to be after Object.assign so an undefined 'tags' field doesn't overwrite the default []
+        this.tags = props.tags || [];
+        if (props.include) this.tags.push("external JS");
```
CharlieS1103, line 87: Capital "E" maybe? Just a preference though

src/Card.js
```diff
@@ -372,6 +383,49 @@ class Card extends react.Component {
         return authorsDiv;
     }
 
+    generateTags(tags) {
+        return tags.reduce((accum, tag) => {
+            // Render tags if enabled. Always render external JS tag
+            if (CONFIG.visual.tags || tag === "external JS") {
```
CharlieS1103, line 389: Remember to Capitalize here as well if you decide to go ahead with that change

CharlieS1103 (approved):

index.js
theRealPadster: Hmm, maybe. Then we'd have to parse the entire thing if we want to pull one out (for example, to see if stars are enabled in Card.js), but that's not super difficult. I'd suggest adding a new issue or PR for that if you like since it's not super related to this functionality.

CharlieS1103: Will do

PR #181, opened 2022-05-02
feat(index): add search function

kyrie25:
- Adds search function for Marketplace (or at least finishes it), thought this was much needed
Currently compare search query with name/user and filters array
Spotify_cDEYQ9TSfY

- Line 843: Unnecessary condition ig?
image

### DIFF
--- a/index.js
+++ b/index.js
@@ -164,6 +164,8 @@ class Grid extends react.Component {
             activeScheme: CONFIG.theme.activeScheme,
             activeThemeKey: CONFIG.theme.activeThemeKey,
         };
+
+        this.handleSearch = this.handleSearch.bind(this);
     }
 
     // TODO: should I put this in Grid state?
@@ -518,6 +520,25 @@ class Grid extends react.Component {
         return this.state.activeScheme;
     }
 
+    /**
+     * @param {import("react").KeyboardEvent} event
+     */
+    handleSearch (event) {
+        if (event.key === "Enter") {
+            this.setState({ endOfList: false });
+            this.newRequest(ITEMS_PER_REQUEST);
+            requested = true;
+        } else if ( // Refreshes result when user deletes all queries
+            ((event.key === "Backspace") || (event.key === "Delete")) &&
+            requested &&
+            this.state.searchValue.trim() === ""
+        ) {
+            this.setState({ endOfList: false });
+            this.newRequest(ITEMS_PER_REQUEST);
+            requested = false;
+        }
+    }
+
     render() {
         return react.createElement("section", {
             className: "contentSpacing",
@@ -557,23 +578,7 @@ class Grid extends react.Component {
                 this.setState({ searchValue: event.target.value });
                 searchQuery = event.target.value;
             },
-            onKeyDown: (event) => {
-                if (event.key === "Enter") {
-                    this.setState({ endOfList: false });
-                    endOfList = false;
-                    this.newRequest(ITEMS_PER_REQUEST);
-                    requested = true;
-                } else if ( // Refreshes result when user deletes all queries
-                    ((event.key === "Backspace") || (event.key === "Delete")) &&
-                    requested &&
-                    this.state.searchValue.trim() === ""
-                ) {
-                    this.setState({ endOfList: false });
-                    endOfList = false;
-                    this.newRequest(ITEMS_PER_REQUEST);
-                    requested = false;
-                }
-            },
+            onKeyDown: (event) => this.handleSearch(event),
         })),
         ),
         [ // Add a header and grid for each card type if it has any cards
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 /// <reference types="react-dom" />
 /// <reference path="../spicetify-cli/globals.d.ts" />
 /// <reference path="../spicetify-cli/jsHelper/spicetifyWrapper.js" />
+/// <reference path="src/AddSnippetModal.js" />
 /// <reference path="src/Card.js" />
 /// <reference path="src/Icons.js" />
 /// <reference path="src/Settings.js" />
@@ -646,6 +647,7 @@ async function getExtensionRepos(page = 1) {
     const allRepos = await fetch(url).then(res => res.json()).catch(() => []);
     if (!allRepos.items) {
         Spicetify.showNotification("Too Many Requests, Cool Down.");
+        return null;
     }
     const filteredResults = {
         ...allRepos,
@@ -841,6 +843,7 @@ async function getThemeRepos(page = 1) {
     const allThemes = await fetch(url).then(res => res.json()).catch(() => []);
     if (!allThemes.items) {
         Spicetify.showNotification("Too Many Requests, Cool Down.");
+        return null;
     }
     const filteredResults = {
         ...allThemes,
--- a/index.js
+++ b/index.js
@@ -141,6 +141,7 @@ let requestPage = null;
 const ITEMS_PER_REQUEST = 100;
 
 let BLACKLIST = [];
+let searchQuery = "";
 
 // eslint-disable-next-line no-redeclare, no-unused-vars
 let gridUpdateTabs, gridUpdatePostsVisual;
@@ -249,7 +250,8 @@ class Grid extends react.Component {
     // Returns the next page number to fetch, or null if at end
     // TODO: maybe we should rename `loadPage()`, since it's slightly confusing when we have github pages as well
     async loadPage(queue) {
-        if (CONFIG.activeTab === "Extensions") {
+        switch (CONFIG.activeTab) {
+        case "Extensions": {
             let pageOfRepos = await getExtensionRepos(requestPage);
             for (const repo of pageOfRepos.items) {
                 let extensions = await fetchExtensionManifest(repo.contents_url, repo.default_branch, repo.stargazers_count);
@@ -276,7 +278,8 @@ class Grid extends react.Component {
             console.log(`Parsed ${soFarResults}/${pageOfRepos.total_count} extensions`);
             if (remainingResults > 0) return currentPage + 1;
             else console.log("No more extension results");
-        } else if (CONFIG.activeTab === "Installed") {
+            break;
+        } case "Installed": {
             const installedStuff = {
                 theme: getLocalStorageDataFromKey(LOCALSTORAGE_KEYS.installedThemes, []),
                 extension: getLocalStorageDataFromKey(LOCALSTORAGE_KEYS.installedExtensions, []),
@@ -299,10 +302,10 @@ class Grid extends react.Component {
                     });
                 }
             }
-
+            break;
             // Don't need to return a page number because
             // installed extension do them all in one go, since it's local
-        } else if (CONFIG.activeTab == "Themes") {
+        } case "Themes": {
             let pageOfRepos = await getThemeRepos(requestPage);
             for (const repo of pageOfRepos.items) {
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -537,21 +538,34 @@ class Grid extends react.Component {
...
+                    const { searchValue } = this.state;
+                    const { title, user } = card.props;
+
+                    if (searchValue.trim() === "")
+                        return card;
+                    else if (
+                        title.toLowerCase().includes(searchValue.trim().toLowerCase()) ||
+                        user.toLowerCase().includes(searchValue.trim().toLowerCase())
+                    )
+                        return card;
```
theRealPadster, line 567:
```suggestion
                    if (
                        searchValue.trim() === "" ||
                        title.toLowerCase().includes(searchValue.trim().toLowerCase()) ||
                        user.toLowerCase().includes(searchValue.trim().toLowerCase())
                    ) return card;
```
Can merge together

index.js
```diff
@@ -826,7 +840,7 @@ async function getThemeRepos(page = 1) {
     return filteredResults;
 }
 function addToSessionStorage(items, key) {
-    if (!items || items == null) return;
+    if (!items) return;
```
theRealPadster, line 843: Good call lol

theRealPadster (changes requested):
Hey thanks for tackling this! Looks good, just one suggestion.  My other concern is the fact that we're using paging and caching now. I was thinking the search filter would send an actual search query to github and fetch results frrom there. If we just filter what's in the cards variable, it could miss things if they haven't all been loaded. Not sure if it makes sense to add in the meantime (since it could lead to bug reports about the search missing things). Thoughts @CharlieS1103 ?

index.js
kyrie25: Oh yeah that part came out different on my head lmao, thanks for calling that out.

kyrie25:
> If we just filter what's in the cards variable, it could miss things if they haven't all been loaded.

I'm not sure it would be that much of an issue since all the GitHub results are loaded almost instantly every time Spotify starts (at least for me) but yeah that does make sense
image

kyrie25:
Also I've found out that sending requests continuously to GitHub's Search API [can eventually lead to a bottleneck](https://docs.github.com/en/rest/search#timeouts-and-incomplete-results) as `onChange` is fired on every new input so ig we'll have to implement an event that sends a request after `onChange` has been fired for a certain period of time?
> Timeouts and incomplete results
To keep the Search API fast for everyone, we limit how long any individual query can run. For queries that [exceed the time limit](https://developer.github.com/changes/2014-04-07-understanding-search-results-and-potential-timeouts/), the API returns the matches that were already found prior to the timeout, and the response has the incomplete_results property set to true.
Reaching a timeout does not necessarily mean that search results are incomplete. More results might have been found, but also might not.

CharlieS1103 (approved):
Thanks for doing this! Seems all good to me. One functionality we may want to add in the future is searching by tag.

CharlieS1103:
Just took a second look and seems like we may run into issues using github requests, might need to migrate the onchange to use the caching system i implemented

kyrie25:
> Or use a hybrid where we search the cards like you're currently doing sometimes maybe 🤔

I'm pretty sure if we were to use GitHub API for search queries it wouldn't be much of a problem to implement a hybrid system as this is possible
```js
if (allThemes.incomplete_results === true) {
        // Filter results from cache
    }
```
I'm not sure about how to deal with the throttle part tho, gonna have to look into it a bit more.

kyrie25:
> Just took a second look and seems like we may run into issues using github requests, might need to migrate the onchange to use the caching system i implemented

If I may ask could you point me to that system so I can take a look there

kyrie25:
I've found out also that if you were to send a incomplete keyword (such as `play` instead of `playlist`) the result would return _fewer results_ than the actual keyword (even though it still includes `playlist` results), and that it would be difficult to differentiate  between a repository/author/tag/etc. query, apart from having users write `user:` themselves, so I'm not sure there's any possible way to implement live GitHub API results and maintaining a good UX at the moment
image
image

kyrie25:
> Just took a second look and seems like we may run into issues using github requests, might need to migrate the onchange to use the caching system i implemented

Oh I have yet to implement GitHub API calls for search queries so I don't think that's a problem yet

theRealPadster:
> I've found out also that if you were to send an incomplete keyword (such as `play` instead of `playlist`) the API would return _fewer results_ than the actual keyword (even though it still includes `playlist` results)

Woah, that is bizarre. 
I did this search of [play](https://api.github.com/search/repositories?q=play+topic%3Aspicetify-extensions&per_page=100) vs [playist](https://api.github.com/search/repositories?q=playlist+topic%3Aspicetify-extensions&per_page=100). It looks to me like the ones that don't show up for "play" but do for "playlist" have the word "playlist" in text fields or tags, and nowhere in any URLs. Maybe it tries to match full words but just accepts full url strings? 

I think the best solution for rate limits etc would be to actually just not have live search and to have it load a search page when you hit enter. That solves the query issue, and people will type the entire thing in. 

We could also do your card filtering method for live results, and then have a full search page if they press enter. Thoughts?

kyrie25:
> I think the best solution for rate limits etc would be to actually just not have live search and to have it load a search page when you hit enter. That solves the query issue, and people will type the entire thing in.
> 
> We could also do your card filtering method for live results, and then have a full search page if they press enter. Thoughts?

That is a pretty good idea, I agree. I was thinking about caching the initial results also, and if GitHub returns 0 results we could filter out from the cache and give the results from there.

Oh and please feel free to commit directly to the PR once you find a working solution :) . I might not be able to find one in a timely manner (mainly because this is the first time I've worked with React) so I'd appreciate any help also.

kyrie25:
Now filters existing results upon input, fetch live GitHub API results (aka. refresh results) upon pressing Enter key
Spotify_iLN0jGxYKJ

kyrie25:
Addresses some small safety checks
image

theRealPadster:
Hey sorry, I've been super busy lately. Really do appreciate the contributions though. I'll look as soon as I can. Wanna get the code running locally.

kyrie25:
Now uses loading icon when fetching queries
Spotify_bizwOWy8SX

index.js
```diff
@@ -537,21 +546,52 @@ class Grid extends react.Component {
...
+                    endOfList = false;
+                    this.newRequest(ITEMS_PER_REQUEST);
+                    requested = true;
+                } else if ( // Refreshes result when user deletes all queries
+                    ((event.key === "Backspace") || (event.key === "Delete")) &&
+                    requested &&
+                    this.state.searchValue.trim() === ""
+                ) {
+                    this.setState({ endOfList: false });
+                    endOfList = false;
```
theRealPadster, line 572: I'm noticing a local variable and state variable for `endOfList` in a bunch of places. Is there a reason why?

index.js
```diff
@@ -610,7 +650,10 @@ class Grid extends react.Component {
  */
 async function getExtensionRepos(page = 1) {
     // www is needed or it will block with "cross-origin" error.
-    let url = `https://api.github.com/search/repositories?q=${encodeURIComponent("topic:spicetify-extensions")}&per_page=${ITEMS_PER_REQUEST}`;
+    let url = `https://api.github.com/search/repositories?q=${encodeURIComponent(`topic:spicetify-extensions`)}&per_page=${ITEMS_PER_REQUEST}`;
```
theRealPadster, line 653: Not a big fan of string templates when there's no interpolation used. 
```suggestion
    let url = `https://api.github.com/search/repositories?q=${encodeURIComponent("topic:spicetify-extensions")}&per_page=${ITEMS_PER_REQUEST}`;
```

index.js
```diff
@@ -610,7 +650,10 @@ class Grid extends react.Component {
...
+    if (searchQuery?.trim() !== "") {
+        url = `https://api.github.com/search/repositories?q=${encodeURIComponent(`${searchQuery} topic:spicetify-extensions`)}&per_page=${ITEMS_PER_REQUEST}`;
```
theRealPadster, line 655: This would work as well, right?
```suggestion
        url = `https://api.github.com/search/repositories?q=${encodeURIComponent(`${searchQuery}+topic:spicetify-extensions`)}&per_page=${ITEMS_PER_REQUEST}`;

```

index.js
```diff
@@ -804,7 +848,9 @@ async function fetchThemeManifest(contents_url, branch, stars) {
  */
 async function getThemeRepos(page = 1) {
     let url = `https://api.github.com/search/repositories?q=${encodeURIComponent("topic:spicetify-themes")}&per_page=${ITEMS_PER_REQUEST}`;
-
+    if (searchQuery?.trim() !== "") {
+        url = `https://api.github.com/search/repositories?q=${encodeURIComponent(`${searchQuery} topic:spicetify-themes`)}&per_page=${ITEMS_PER_REQUEST}`;
```
theRealPadster, line 852:
```suggestion
        url = `https://api.github.com/search/repositories?q=${encodeURIComponent(`${searchQuery}+topic:spicetify-themes`)}&per_page=${ITEMS_PER_REQUEST}`;
```

index.js
```diff
@@ -804,7 +848,9 @@ async function fetchThemeManifest(contents_url, branch, stars) {
  */
 async function getThemeRepos(page = 1) {
```
theRealPadster, line 849: Is it possible to pass the search query into the function? And same for the "get extensions" one. Would make for more readable code.

theRealPadster (changes requested):
Loading icon is sweet! I have a couple suggestions. I'd also like it mention that we are (I am) in the process of converting the app to be built with Spicetify-Creator, which lets us use jsx and typescript and other niceties. So these changes will need to be replicated once that is ready. Do you think you'd be able to get the functionality moved over once we've moved?

index.js
kyrie25: I used it to trigger the loading icon in my latest commit. Otherwise, I'm pretty sure it was implemented before I made this PR? So I'm not entirely sure.

index.js
kyrie25: I'll look into it ASAP

kyrie25:
> Loading icon is sweet! I have a couple suggestions. I'd also like it mention that we are (I am) in the process of converting the app to be built with Spicetify-Creator, which lets us use jsx and typescript and other niceties. So these changes will need to be replicated once that is ready. Do you think you'd be able to get the functionality moved over once we've moved? 

Sure, I'll look into it! Do give me a mention on GitHub or Discord once you need it going.

index.js
kyrie25: It does yeah, gonna apply suggestion.

index.js
kyrie25: I'm not sure it is possible, more accurately it _is_ possible, but a new request needs to be called (calling the `newRequest()` function), so in order to use search queries as parameters, I would have to nest the variable through:
```js
newRequest() => loadAmount() => loadPage() => getThemeRepos()
```
Which I'm pretty certain won't do any good in refactoring.

index.js
kyrie25: Worked without reassigning local variable.

index.js
theRealPadster: Ah, yeah, that's fair. Some of the legacy code is a little monolithic.

index.js
theRealPadster: Oh, right, yeah, it's one of those globals that were just sitting around. Glad it works with the state. Makes me wonder why it has a local at all though lol. We can figure out later.

CharlieS1103:
Going to go ahead and just merge this in, we'll get it moved over to JSX when the time comes.

PR #213, opened 2022-06-04
Spicetify Creator refactor

theRealPadster:
Still need to update the install scripts, and the actions.


theRealPadster:
@afonsojramos You use Windows, right? Could you take a look at or test the install script?

resources/install.ps1
```diff
@@ -24,18 +24,18 @@ if (-not (Test-Path $sp_dot_dir)) {
...
+Expand-Archive -Path "${HOME}/spicetify-cli/CustomApps/marketplace.zip" -DestinationPath "${HOME}/spicetify-cli/CustomApps/" -Force
+Remove-Item -Path "${HOME}/spicetify-cli/CustomApps/marketplace.zip" -Force
+if (Test-Path -Path "${HOME}/spicetify-cli/CustomApps/marketplace") {
+  Write-Host "marketplace was already found! Updating..." -ForegroundColor "Cyan"
+  Remove-Item -Path "${HOME}/spicetify-cli/CustomApps/marketplace" -Force -Recurse
 }
-Rename-Item -Path "${HOME}/spicetify-cli/CustomApps/spicetify-marketplace-main" -NewName "spicetify-marketplace" -Force
-Copy-Item -Path "${HOME}/spicetify-cli/CustomApps/spicetify-marketplace" -Destination $sp_dot_dir -Recurse -Force
-spicetify config custom_apps spicetify-marketplace
+Rename-Item -Path "${HOME}/spicetify-cli/CustomApps/marketplace-dist" -NewName "marketplace" -Force
```
kyrie25, line 36:
```suggestion
Rename-Item -Path "${HOME}/spicetify-cli/CustomApps/spicetify-marketplace-dist" -NewName "marketplace" -Force
```

kyrie25 (changes requested):
Thrown an error due to wrong folder name
image

resources/install.ps1
```diff
@@ -24,18 +24,18 @@ if (-not (Test-Path $sp_dot_dir)) {
...
+Copy-Item -Path "${HOME}/spicetify-cli/CustomApps/marketplace" -Destination $sp_dot_dir -Recurse -Force
+spicetify config custom_apps marketplace
```
kyrie25, line 38:
```suggestion
spicetify config custom_apps spicetify-marketplace-
spicetify config custom_apps marketplace
```
Remove the old marketplace from file as it applies both if it is previously installed
image

resources/install.ps1
```diff
@@ -24,18 +24,18 @@ if (-not (Test-Path $sp_dot_dir)) {
...
 spicetify backup apply
```
kyrie25, line 39:
```suggestion
spicetify backup
spicetify apply
```
Split commands as Spicetify would not let you apply if it is stacked right after `backup`, thus making the user manually applying again.

kyrie25 (changes requested):
Few improvements as well

resources/install.ps1
CharlieS1103: Are you sure about this? I'm pretty sure it's worked in the past
ex: ```spicetify restore backup apply```

kyrie25: It works with `spicetify restore backup apply` (the whole chain), not `spicetify backup apply` which would just stop at `backup`

kyrie25: But using `restore backup apply` would take a significant amount of time plus most of the case Marketplace is installed right after the user installs Spicetify (bundled in docs) which is unnecessary.

afonsojramos: Maybe this is a bug in spicetify cli, but I've also had the issue detailed above.

theRealPadster:
Thanks @kyrie25! Did I miss anything?

kyrie25:
> Thanks @kyrie25! Did I miss anything?

So far so good yeah!

src/components/Grid.tsx
```diff
@@ -0,0 +1,521 @@
...
+      schemes: props.CONFIG.theme.schemes,
+      activeScheme: props.CONFIG.theme.activeScheme,
+      activeThemeKey: props.CONFIG.theme.activeThemeKey,
+    };
+  }
+
+  searchRequested: boolean;
+  endOfList = false;
+  lastScroll = 0;
+  requestQueue: any[] = [];
```
kyrie25, line 62:
```suggestion
  requestQueue: never[][] = [];
```
Not important but if you want to get rid of as most implicit `any` as possible

PR #220, opened 2022-06-08
fix(Card): use snippet URL if card is a snippet

kyrie25:
Fix `href` for snippet Cards as they don't have their own repos and authors field and would return `undefined`

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -415,7 +415,7 @@ export default class Card extends React.Component<CardProps, {
               title={this.props.type === "snippet" ? this.props.item.title : this.props.item.manifest?.name}
               className="main-cardHeader-link"
               dir="auto"
-              href={this.props.type === "snippet"
+              href={this.props.type !== "snippet"
                 ? `https://github.com/${this.props.item.user}/${this.props.item.repo}`
                 : "https://github.com/spicetify/spicetify-marketplace/blob/main/resources/snippets.json"
               }
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -415,7 +415,10 @@ export default class Card extends React.Component<CardProps, {
               title={this.props.type === "snippet" ? this.props.item.title : this.props.item.manifest?.name}
               className="main-cardHeader-link"
               dir="auto"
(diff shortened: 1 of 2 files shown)


theRealPadster:
Can you send a screenshot of what this looks like? Does it render a page inside Spotify with the snippets json?

src/components/Card/Card.tsx
```diff
@@ -415,7 +415,10 @@ export default class Card extends React.Component<CardProps, {
               title={this.props.type === "snippet" ? this.props.item.title : this.props.item.manifest?.name}
               className="main-cardHeader-link"
               dir="auto"
-              href={`https://github.com/${this.props.item.user}/${this.props.item.repo}`}
+              href={this.props.type !== "snippet"
+                ? `https://github.com/${this.props.item.user}/${this.props.item.repo}`
+                : "https://github.com/spicetify/spicetify-marketplace/blob/main/resources/snippets.json"
```
CharlieS1103, line 420: You can use the SNIPPETS_URL constant rather than hard coding the URL in here

CharlieS1103 (changes requested):

src/components/Card/Card.tsx
kyrie25: SNIPPETS_URL returns a Raw GirHub link, I'm trying to use the Web UI one.

kyrie25:
> Can you send a screenshot of what this looks like? Does it render a page inside Spotify with the snippets json?

It's just the href used on the title of the cards that would open a new browser window on click.

CharlieS1103 (approved):
All good to me!

theRealPadster:
> > Can you send a screenshot of what this looks like? Does it render a page inside Spotify with the snippets json?
> 
> It's just the href used on the title of the cards that would open a new browser window on click.

Ohh right, that's perfect. My only suggestion would be to add a SNIPPETS_PAGE_URL constant and use that.

kyrie25:
> > It's just the href used on the title of the cards that would open a new browser window on click.
> 
> Ohh right, that's perfect. My only suggestion would be to add a SNIPPETS_PAGE_URL constant and use that.

Yeah sure, I can do that if you prefer so.

theRealPadster:
Thanks! Merged 👌

PR #226, opened 2022-06-09
✨➕ Add syntax highlighting and image previews to snippets

theRealPadster:
Added syntax highlighting to snippet modals via [react-simple-code-editor](https://www.npmjs.com/package/react-simple-code-editor). I needed to add a script tag to fix a bug with it (in extension.ts), and for some reason it wouldn't build if I imported the CSS from node_modules, so I copied the theme file into the app. 

I've also added in the changes to add preview images for snippets, and the capability to add your own, saved as base64 in localstorage.


kyrie25 (approved):
Works perfectly fine for me!

PR #231, opened 2022-06-11
fix: several changes

### LINKED ISSUE #196: Upgrade button showing every time Spotify starts although the snippet to hide it is installed
### 🔍 Have you checked Spicetify.app page for your issue?

- [X] I have checked the FAQ

### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version: Spotify for Windows 1.1.86.857.g3751ee08
- Spicetify version: 2.10.1
- Windows 10 21H2
```

### 📝 Description

Every time Spotify starts, I have to reinstall the Hide Upgrade Button Snippet to get it working. This problem is not new to this version, as this has been a problem since late April (it was Spotify 1.184.716 and Spicetify 2.9.8 iirc).

### 📸 Screenshots

image

kyrie25:
Resolves #196 
Changes derived from #229, previews are stored there

### DIFF
--- a/src/components/ReadmePage.tsx
+++ b/src/components/ReadmePage.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import LoadingIcon from "./Icons/LoadingIcon";
 
 class ReadmePage extends React.Component<
 {
@@ -86,7 +87,9 @@ class ReadmePage extends React.Component<
         <div className="marketplace-header">
           <h1>{this.props.title}</h1>
         </div>
-        <div id="marketplace-readme" className="marketplace-readme__container" dangerouslySetInnerHTML={{ __html: this.state.html }}></div>
+        {this.state.html === "<p>Loading...</p>"
+          ? <footer className="marketplace-footer"><LoadingIcon /></footer>
+          : <div id="marketplace-readme" className="marketplace-readme__container" dangerouslySetInnerHTML={{ __html: this.state.html }}></div>}
       </section>
     );
   }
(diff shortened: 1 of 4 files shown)


src/components/ReadmePage.tsx
```diff
@@ -86,7 +86,9 @@ class ReadmePage extends React.Component<
         <div className="marketplace-header">
           <h1>{this.props.title}</h1>
         </div>
-        <div id="marketplace-readme" className="marketplace-readme__container" dangerouslySetInnerHTML={{ __html: this.state.html }}></div>
+        {this.state.html === "<p>Loading...</p>"
+          ? <footer className="marketplace-footer"><LoadingIcon /></footer>
+          : <div id="marketplace-readme" className="marketplace-readme__container" dangerouslySetInnerHTML={{ __html: this.state.html }}></div>}
```
theRealPadster, line 91: I wonder if it's possible to just make the `state.html` the loading markup. Or maybe not because it's JSX and the html needs to be a string?

theRealPadster (approved):

src/components/ReadmePage.tsx
kyrie25: I tried it before, but it just returns `[object Object]` the exact string, sometimes it won't even update.

theRealPadster: Ah yeah, that makes sense. It's trying to convert the loading icon component to a string, so the best it can do is output that it's an object. Makes sense. I briefly looked into if we could make the page itself JSX, but it doesn't look like that's super easy or worth it.

PR #234, opened 2022-06-11
feat(Grid): add update checker

kyrie25:
- Feature update:
Now checks for new updates every time Marketplace is opened and notifies the user via a button that leads to the repository:
image

- Styling update:
Merge search bar onto header

- TODO:
- [x] Use an SVG instead of `available!` for styling purposes(?)
- [x] Implement [spcr-whats-new](https://github.com/jeroentvb/spcr-whats-new) and manage changelogs

### DIFF
--- a/src/components/Grid.tsx
+++ b/src/components/Grid.tsx
@@ -424,13 +424,8 @@ export default class Grid extends React.Component<
     return this.state.activeScheme;
   }
 
-  changelogModal() {
-    return Changelog;
-  }
-
   render() {
     return (
-
       <section className="contentSpacing">
         <div className="marketplace-header">
           <div className="marketplace-header__left">
--- a/src/components/Grid.tsx
+++ b/src/components/Grid.tsx
@@ -1,7 +1,8 @@
 import React from "react";
+import semver from "semver";
 import { CardItem, CardType, Config, SchemeIni, Snippet, TabItemConfig, TabType } from "../types/marketplace-types";
 import { getLocalStorageDataFromKey, generateSchemesOptions, injectColourScheme } from "../logic/Utils";
-import { LOCALSTORAGE_KEYS, ITEMS_PER_REQUEST, MARKETPLACE_VERSION, REMOTE_VERSION } from "../constants";
+import { LOCALSTORAGE_KEYS, ITEMS_PER_REQUEST, MARKETPLACE_VERSION, LATEST_RELEASE } from "../constants";
 import { openModal } from "../logic/LaunchModals";
 import {
   getExtensionRepos, fetchExtensionManifest,
@@ -354,14 +355,14 @@ export default class Grid extends React.Component<
   */
   async componentDidMount() {
     // Checks for new Marketplace updates
-    fetch(REMOTE_VERSION).then(res => res.json()).then(
+    fetch(LATEST_RELEASE).then(res => res.json()).then(
       result => {
         this.setState({
-          version: result.version,
+          version: result[0].name,
         });
       },
       error => {
-        console.log("Failed to check for updates", error);
+        console.error("Failed to check for updates", error);
       },
     );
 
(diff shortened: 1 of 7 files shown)


kyrie25:
Implemented `spcr-whats-new` to display a changelog modal every time it is updated.
image

**Important note**: Current version of Marketplace is **not** a valid SemVer version, so in order to trigger the modal, the version inside the Changelog component **must** be bumped simultaneously as the version number of Marketplace.
image

src/components/Grid.tsx
```diff
@@ -349,6 +353,20 @@ export default class Grid extends React.Component<
   * If the cardList isn't loaded, it loads the cardList.
   */
   async componentDidMount() {
+    // Checks for new Marketplace updates
+    fetch("https://raw.githubusercontent.com/spicetify/spicetify-marketplace/main/package.json").then(res => res.json()).then(
```
theRealPadster, line 357: Can we add a const for this as well?

src/components/Grid.tsx
```diff
@@ -406,11 +424,26 @@ export default class Grid extends React.Component<
     return this.state.activeScheme;
   }
 
+  changelogModal() {
```
theRealPadster, line 427: Is this used anywhere?

kyrie25: Yeah it was for testing purposes, cleaned it up afterwards.

src/components/Grid.tsx
kyrie25: Definitely yeah, will add ASAP.

kyrie25: Done.

kyrie25:
Now uses GitHub commit log as changelog. Will need to squash commits in the future to reduce list length.
image

src/components/Modals/Changelog/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <a href="https://github.com/spicetify/spicetify-marketplace/pull/226">(#226)</a>
+      </li>
+      <li>General improvements</li>
+    </ul>
+  </>
+); */
+const commitArray: string[] = [];
+
+const changelogDetails = () => {
+  fetch(COMMIT_LIST).then(res => res.json())
```
theRealPadster, line 33: I think you need to `await` this, since fetch is a promise.

src/constants.ts
```diff
@@ -38,3 +38,9 @@ export const SNIPPETS_URL = "https://raw.githubusercontent.com/spicetify/spiceti
 export const SNIPPETS_PAGE_URL = "https://github.com/spicetify/spicetify-marketplace/blob/main/resources/snippets.json";
 
 export const BLACKLIST_URL = "https://raw.githubusercontent.com/spicetify/spicetify-marketplace/main/resources/blacklist.json";
+
+export const REMOTE_VERSION = "https://raw.githubusercontent.com/spicetify/spicetify-marketplace/main/package.json";
+
+// Used for Changelog
+const commitTimestamp = "2022-06-07T21:05:22Z"; // Must be in ISO format
```
theRealPadster, line 45: Will this need to be updated whenever we bump a version as well? I was thinking we do the commits instead of manually specifying changelog notes to save us from having the change things in a bunch of places. Is it possible to just grab the latest release from GitHub and use the notes from that? I'm also realizing we probably don't want to notify people every time we bump the version without a release, since I've been trying to version bump more often when we add a new feature or something. Or we do releases more frequently? Hope this makes some sense, just kind of thinking out loud.

kyrie25: It does, yeah.
I also thought about fetching changelogs and release notes from GitHub but we barely push or publish there (already skipped 0.6.0 and 0.6.1) so it didn't seem plausible for me.

src/components/Modals/Changelog/index.tsx
kyrie25: Good catch. Applied suggestion.

src/constants.ts
kyrie25: If you do decide to go along with publishing releases more frequently then I have already implemented the feature just in case.
image

theRealPadster (approved):

PR #235, opened 2022-06-12
fix(TabRow): Marketplace crashing on moving tabs

kyrie25:
Fixed an issue where Marketplace would crash on moving tabs to non-existent index

<details>
<summary>Details</summary>

- Before:
Spotify_cGbxTCwC9i

- After:
Spotify_duNqPErbVF

</details>

### DIFF
--- a/src/components/Modals/Settings/TabRow.tsx
+++ b/src/components/Modals/Settings/TabRow.tsx
@@ -58,12 +58,12 @@ const TabRow = (props: {
     <div className="setting-row">
       <label htmlFor={toggleId} className='col description'>{props.name}</label>
       <div className="col action">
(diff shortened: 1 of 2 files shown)


theRealPadster (changes requested):
This would probably be better with a disabled prop passed in, and just setting it to disabled for the first and last tabrow, so it won't be hard coded into the third one if we add more in the future

kyrie25:
> This would probably be better with a disabled prop passed in, and just setting it to disabled for the first and last tabrow, so it won't be hard coded into the third one if we add more in the future

I'm pretty sure there's no need to change anything for the first row then, will look into the last row.

theRealPadster (approved):
Ah, perfect, that's better

PR #237, opened 2022-06-15
fix: theme color injection workaround

kyrie25:
Workaround for the color injection problem that can be encountered when installing themes through Marketplace when no local theme is applied.
Also added the `preview` key for the Snippets part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # spicetify-marketplace
 
-Download extensions and themes directly from within [Spicetify](https://github.com/spicetify/spicetify-cli). 
+Download extensions and themes directly from within [Spicetify](https://github.com/spicetify/spicetify-cli).
 
 Made by [@theRealPadster](https://github.com/theRealPadster) and [@CharlieS1103](https://github.com/CharlieS1103).
 
-Based on the [reddit Custom App](https://github.com/spicetify/spicetify-cli/wiki/Custom-Apps#reddit)  
+Based on the [reddit Custom App](https://github.com/spicetify/spicetify-cli/wiki/Custom-Apps#reddit)
 Made with [Spicetify Creator](https://github.com/spicetify/spicetify-creator)
 
 <p align="center">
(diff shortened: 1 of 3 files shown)


resources/install.sh
```diff
@@ -34,6 +34,12 @@ cd "$INSTALL_DIR/marketplace"
 # Remove old custom app name if exists
 spicetify config custom_apps spicetify-marketplace-
 
+# Color injection fix
+spicetify config inject_css 1
+spicetify config replace_colors 1
+spicetify config current_theme SpicetifyDefault
+spicetify config color_scheme SpicetifyDefault
```
theRealPadster, line 41: Is this a thing? Mine worked with the colour scheme blank

resources/install.sh
```diff
@@ -34,6 +34,12 @@ cd "$INSTALL_DIR/marketplace"
 # Remove old custom app name if exists
 spicetify config custom_apps spicetify-marketplace-
 
+# Color injection fix
+spicetify config inject_css 1
+spicetify config replace_colors 1
+spicetify config current_theme SpicetifyDefault
```
theRealPadster, line 40: Is there a way to check if they have the colour scheme set? People will get confused if they install marketplace and then lose their theme.

README.md
```diff
@@ -1,10 +1,10 @@
 # spicetify-marketplace
 
-Download extensions and themes directly from within [Spicetify](https://github.com/spicetify/spicetify-cli). 
+Download extensions and themes directly from within [Spicetify](https://github.com/spicetify/spicetify-cli).
 
 Made by [@theRealPadster](https://github.com/theRealPadster) and [@CharlieS1103](https://github.com/CharlieS1103).
 
-Based on the [reddit Custom App](https://github.com/spicetify/spicetify-cli/wiki/Custom-Apps#reddit)  
+Based on the [reddit Custom App](https://github.com/spicetify/spicetify-cli/wiki/Custom-Apps#reddit)
```
theRealPadster, line 7: Can we remove this? I feel like it's diverged enough and there's not really much point adding this.

CharlieS1103: I agree with this @theRealPadster , plus lyrics-plus and new-releases both use the same system

README.md
```diff
@@ -1,10 +1,10 @@
 # spicetify-marketplace
 
-Download extensions and themes directly from within [Spicetify](https://github.com/spicetify/spicetify-cli). 
+Download extensions and themes directly from within [Spicetify](https://github.com/spicetify/spicetify-cli).
 
 Made by [@theRealPadster](https://github.com/theRealPadster) and [@CharlieS1103](https://github.com/CharlieS1103).
```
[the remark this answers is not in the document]
CharlieS1103: Agreed

resources/install.sh
kyrie25: Seems like it is possible, will patch in a sec.

CharlieS1103: @kyrie25 I don't run windows unfortunately.

theRealPadster (approved):
I think this is good, right? The bash syntax checks out

kyrie25:
Don't think there are anything else to do. If the bash script works then it should be ready to merge.

PR #241, opened 2022-06-18
fix(Card): remove duplicate tags from appearing

kyrie25:
Fix an issue where the Card tags could duplicate when rendered in `Installed` tab

image
image

Also declared types for ease of maintenance.


kyrie25:
I had no idea why it had duplicate tags in the first place but the size of the array seems small enough so the performance overhead won’t be noticeable.

theRealPadster:
> I had no idea why it had duplicate tags in the first place but the size of the array seems small enough so the performance overhead won’t be noticeable.

Fair point lol. I think the filter beforehand is more efficient than anything I could come up with inside the reducer.

theRealPadster (approved):

PR #243, opened 2022-06-19
Add custom apps tab

theRealPadster:
Work-in-progress.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ CSS snippets are rather basic to implement. We fetch them from this repo, so you
 1. `componentDidMount` triggers `newRequest`, which triggers `loadAmount(30)`
 2. `loadAmount` calls `loadPage` in a loop until it has the requested amount of cards or runs out of results
 3. `loadPage` calls `getRepos(page)` to get the next page of extensions. It queries the GitHub API for any repos with the "spicetify-extensions" topic. We'll likely add our own tag in the future, like "spicetify-marketplace".
-4. The it loops through all the results and runs `fetchRepoExtensions()` or `getThemeRepos()`, which fetches a `manifest.json` file from the repo's root folder. If it finds one, we generate a card based on the info.
+4. The it loops through all the results and runs `getTaggedRepos()`, which fetches a `manifest.json` file from the repo's root folder. If it finds one, we generate a card based on the info.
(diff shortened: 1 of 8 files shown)


theRealPadster:
Lookin good...

image

README.md
```diff
@@ -174,7 +174,7 @@ CSS snippets are rather basic to implement. We fetch them from this repo, so you
 1. `componentDidMount` triggers `newRequest`, which triggers `loadAmount(30)`
 2. `loadAmount` calls `loadPage` in a loop until it has the requested amount of cards or runs out of results
 3. `loadPage` calls `getRepos(page)` to get the next page of extensions. It queries the GitHub API for any repos with the "spicetify-extensions" topic. We'll likely add our own tag in the future, like "spicetify-marketplace".
-4. The it loops through all the results and runs `fetchRepoExtensions()` or `getThemeRepos()`, which fetches a `manifest.json` file from the repo's root folder. If it finds one, we generate a card based on the info.
+4. The it loops through all the results and runs `getTaggedRepos()`, which fetches a `manifest.json` file from the repo's root folder. If it finds one, we generate a card based on the info.
```
CharlieS1103, line 177: I'm pretty sure that this entire section needs updating since it's no longer as simple as it was before.

src/components/Card/Card.tsx
```diff
@@ -63,6 +64,7 @@ export default class Card extends React.Component<CardProps, {
     if (props.type === "snippet") cardId = props.item.title.replaceAll(" ", "-");
     else if (props.type === "theme") cardId = props.item.manifest?.usercss || "";
     else if (props.type === "extension") cardId = props.item.manifest?.main || "";
+    else if (props.type === "app") cardId = props.item.manifest?.name?.replaceAll(" ", "-") || "";
```
CharlieS1103, line 67: Why are we replacing spaces with hyphens here but no where else?

CharlieS1103 (approved):

src/components/Card/Card.tsx
kyrie25: It is only meant for localStorage as a means to not include spaces in keys

image

kyrie25 (approved):
LGTM

README.md
kyrie25: I'm not completely sure but it seems like it still does the same thing in a nutshell.

Small change needed though
```suggestion
4. Then it loops through all the results and runs `getTaggedRepos()`, which fetches a `manifest.json` file from the repo's root folder. If it finds one, we generate a card based on the info.
```

PR #249, opened 2022-06-22
feat(Changelog & Card): several improvements

### LINKED ISSUE #54: Refresh colour schemes on load?
It downloads the colour schemes on theme install, and caches them with the card/localstorage. While themes themselves will always be up-to-date, since they directly reference the css file, the colour schemes are cached in this way and won't update if the color.ini is updated on GitHub.

### LINKED ISSUE #250: Show a github logo instead of download icon for custom apps
### 📝 Provide a description of the new feature

Show a github logo instead of download for custom apps. The download icon being the same for custom apps and everything else kind of suggests me that once the app is installed, it will change to a trash icon and let you uninstall. This is not possible, and there is no indication if you have the app installed or not. 

### ➕ Additional Information

Thoughts?

### LINKED ISSUE #251: Last updated display
### 📝 Provide a description of the new feature

Show when an extension or theme was last updated. Should help to tell when a theme or extension is outdated.

### ➕ Additional Information

_No response_

kyrie25:
- Use Changelog body rendered from GitHub instead of `react-markdown`
- Add last update section for Cards (resolves #251) and updates extension/themes on new commit (resolves #54)
- Use GitHub icon for Apps (resolves #250)

<details>
<summary>Screenshots</summary>

image
image

</details>


src/components/Modals/Changelog/index.tsx
```diff
@@ -19,18 +17,39 @@ const fetchReleaseBody = async () => {
     });
 };
 
+const getReadmeHTML = async (readmeText: string) => {
```
theRealPadster, line 20: Maybe we should put this in one of the utils files if it's basically the same function as the readme modals?

kyrie25: Makes sense yeah, gonna look into it.

src/components/ReadmePage.tsx
```diff
@@ -25,35 +26,22 @@ class ReadmePage extends React.Component<
...
-      const readmeHtmlRes = await fetch("https://api.github.com/markdown", {
-        method: "POST",
-        body: JSON.stringify(postBody),
-      });
-      if (!readmeHtmlRes.ok) throw Spicetify.showNotification(`Error parsing README (HTTP ${readmeHtmlRes.status})`);
-
-      const readmeHtml = await readmeHtmlRes.text();
-
-      if (readmeHtml == null) {
+  getReadmeHTML = async () => {
```
theRealPadster, line 29: Did you test the functions that I had the "TODO: test this" on?

kyrie25: I tested everything before removing those comments yeah

src/components/Card/Card.tsx
```diff
@@ -489,14 +513,14 @@ export default class Card extends React.Component<CardProps, {
...
-                label={IS_INSTALLED ? "Remove" : "Save"}
+                label={this.props.type === "app" ? "GitHub" : IS_INSTALLED ? "Remove" : "Install"}
                 onClick={(e) => {
                   e.stopPropagation();
                   this.buttonClicked();
                 }}
               >
                 {/*If the extension, theme, or snippet is already installed, it will display trash, otherwise it displays download*/}
-                {IS_INSTALLED ? <TrashIcon /> : <DownloadIcon />}
+                {this.props.type === "app" ? <GitHubIcon /> : IS_INSTALLED ? <TrashIcon /> : <DownloadIcon />}
```
theRealPadster, line 523: At this point I'm wondering if we make a lookup for these.

src/components/Card/Card.tsx
```diff
@@ -118,17 +117,31 @@ export default class Card extends React.Component<CardProps, {
...
+            break;
+          }
         });
       }
 
       if (this.state.lastUpdated !== pushed_at) {
         this.setState({ lastUpdated: pushed_at }, () => {
           console.log(`New update pushed at: ${this.state.lastUpdated}; updating localstorage.`);
-          this.installExtension();
+          switch (this.props.type) {
```
theRealPadster, line 137: Could it ever run both of these? May want to tweak the if statements/logic flow so it can't run on both cases.

kyrie25: I'm pretty sure not, because once the star is updated it would update localStorage also, which makes the new update case false. But we can return the function immediately after updating stars.

src/components/Card/Card.tsx
kyrie25: Will put it on the TODO list

src/components/Card/Card.tsx
kyrie25: Fixed.

PR #256, opened 2022-06-25
Mono-manifest

theRealPadster:
Work-in-progress PR for discussion and so that I can sync main down

TODO:
- [ ]  Get user/repo/branch working (it only passes in the manifest data when using mono-manifest)
- [x]  Need  a repo for custom apps (or add a key to denote app vs extension, and extra logic)
- [ ] Write node.js script to scan items in monomanifest and fetch star counts from github, then update the monomanifest
- [ ] Write a workflow to automate/schedule this
- [ ] Add optionality to extensions (@CharlieS1103 )
- [ ] Add optionality to custom apps (@CharlieS1103 )


src/extensions/extension.tsx
```diff
@@ -194,16 +200,27 @@ async function loadPageRecursive(type: RepoType, pageNum: number) {
   // The recursion isn't super clean...
 
   // TODO: re-enable this once everything works with mono-manifest...
-  // await Promise.all([
-  //   loadPageRecursive("extension", 1),
-  //   loadPageRecursive("theme", 1),
-  // ]);
+   await Promise.all([
```
theRealPadster: We should at least check for the setting here before sending off queries when we're not going to use them

src/logic/FetchTopicRemotes.ts
```diff
@@ -0,0 +1,275 @@
...
+
+// TODO: can we add a return type here?
+/**
+* Fetch extensions from a repo and format data for generating cards
+* @param contents_url The repo's GitHub API contents_url (e.g. "https://api.github.com/repos/theRealPadster/spicetify-hide-podcasts/contents/{+path}")
+* @param branch The repo's default branch (e.g. main or master)
+* @param stars The number of stars the repo has
+* @returns Extension info for card (or null)
+*/
+export async function fetchExtensionManifestFromTopic(contents_url: string, branch: string, stars: number, hideInstalled = false) {
```
theRealPadster, line 76: I'd rather not have so much of this code duplicated with the mono-manifest stuff. That's part of the reason why I didn't really want to support both in the first place.

CharlieS1103: As far as this goes, it more or less just comes down to having two files(which we have) and adding a few if statements in select locations.  I've already worked out the system so there's not much else that needs to be done other than adding extension support to the mono manifest system.

src/styles/components/_dropdown.scss
```diff
@@ -1,6 +1,6 @@
 @use '../constants.scss';
 
-$dropdown-text: rgba(var(--spice-rgb-text), .7);
+$dropdown-text: var(--spice-text);
```
theRealPadster, line 3: It's using the RGB text for a reason here. The opacity is 0.7 because it lights up (to 1.0) on hover. You just removed that functionality.

src/logic/FetchRemotes.ts
```diff
@@ -172,21 +172,14 @@ export function buildAppCardData(manifest: Manifest, user?: string, repo?: strin
...
-    return manifest;
-  } else if (type === "app") {
-    // const manifest = await fetch(APPS_URL).then(res => res.json()).catch(() => null);
-    const manifest = appsManifest;
-    return manifest;
-  } else {
-    return [];
-  }
+  const manifest = {
+    theme: await fetch(THEMES_URL).then(res => res.json()).catch((err) => {console.error(err); return [];}),
```
theRealPadster, line 176: This will make the fetch call every time you run this function, not just for themes.

src/styles/components/_dropdown.scss
CharlieS1103: @theRealPadster Unfortunately a majority of themes don't support RGB text, as far as becoming less opaque on hover, I personally didn't really notice it that much and we can add another effect instead if you'd like. 

As far as the styling goes for the marketplace controls(search bar, scheme selector, settings icon, and theme dev tools button), I think that it'd be best to just use the same color variables so it doesn't look funky on some themes.

theRealPadster: Ah, fair. I didn't realize that people left out the RGB versions. Is it not standard?

kyrie25: As far as it goes, the majority of themes uses color variables and the RGB is only injected where it was originally used on default Spotify, which is not a lot, mostly used for gradients.

rxri:
Closing due to different marketplace behavior in v3 and I'm unsure that this pull request would be merged before v3 release.

PR #259, opened 2022-06-25
chore: resolve type errors

kyrie25:
Resolve type errors present in current build

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -60,4 +60,10 @@ module.exports = {
       "error", { "beforeColon": false, "afterColon": true },
     ],
   },
+  "settings": {
+    "react": {
+      // This is what Spotify uses
+      "version": "16.0",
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -60,4 +60,10 @@ module.exports = {
       "error", { "beforeColon": false, "afterColon": true },
     ],
   },
+  "settings": {
+    "react": {
+      // This is what Spotify uses
+      "version": "16.0",
```
theRealPadster, line 66: image

kyrie25: Interesting, since the version number is what I got from vendor~xpui. Will change.

theRealPadster: Oh weird. And this is from 1.1.86 as well.

theRealPadster (approved):

PR #265, opened 2022-06-30
style: use eslint on marketplace only

rxri:


kyrie25 (approved):
Need to change Husky also as to point it to lint Marketplace only.

PR #277, opened 2022-07-15
feat(ci): add `CODEOWNERS` & cleanup

kyrie25:
Cleanup unnecessary stuff, don't know why it was there.


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @spicetify/marketplace
```
theRealPadster, line 1:
```suggest
* @spicetify/marketplace

```
Newline at EOF

packages/marketplace/src/components/ReadmePage.tsx
```diff
@@ -107,7 +107,6 @@ class ReadmePage extends React.Component<
   }
 
   render() {
-    this.props.data.isInstalled();
```
theRealPadster, line 110: Didn't you add this with the install button thing? I was assuming you needed the code to run for it to render or something, but I guess it's not necessary?

theRealPadster (changes requested):

packages/marketplace/src/components/ReadmePage.tsx
kyrie25: It was when I was testing things, the function returns a boolean value that was assigned as state later. Having this actually breaks Marketplace occasionally.

.github/CODEOWNERS
kyrie25: Will fix later.

theRealPadster (approved):

PR #278, opened 2022-07-16
fix(app): null check for `ReadmePage` & styling

kyrie25:
Apply `margin-left` for every element placed in Heading instead of explicitly declaring them for each element.
Fix current `margin-left` problem also.
image

### DIFF
--- a/packages/marketplace/src/styles/components/_grid.scss
+++ b/packages/marketplace/src/styles/components/_grid.scss
@@ -124,7 +124,8 @@
   flex-grow: 1;
 }
 .searchbar-bar {
-  border: unset;
+  border-style: solid;
+  border-color: var(--spice-sidebar);
   background-color: var(--spice-sidebar) !important;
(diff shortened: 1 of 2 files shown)


packages/marketplace/src/styles/components/_grid.scss
```diff
@@ -19,6 +19,9 @@
 
 .marketplace-header__right, .marketplace-header__left {
   display: flex;
+  & > * {
```
theRealPadster, line 22: Can we change this to `div` or something instead of `*`? So it's not as broad.

theRealPadster (changes requested):

packages/marketplace/src/styles/components/_grid.scss
kyrie25: It is only every direct child element so it's mostly wrappers, but sure.

theRealPadster (approved):

PR #283, opened 2022-07-22
style(grid): topbar margin

kyrie25:
apparently `div` was not inclusive enough
image


theRealPadster (approved):

theRealPadster:
Kinda weird syntax to use but I guess it's cause it's cleaner with the scss nesting?

kyrie25:
> Kinda weird syntax to use but I guess it's cause it's cleaner with the scss nesting?

Yeah, less comma thingy I think.

PR #285, opened 2022-07-22
feat(snippet): add disable homepage recommendation

Tetrax-10:


packages/marketplace/src/resources/snippets.ts
```diff
@@ -77,4 +77,10 @@ export default [
     "code": ".main-rootlist-rootlistItemOverlay {\n  display: none;\n}",
     "preview": "resources/assets/snippets/fix-playlist-hover.png",
   },
+  {
+    "title": "Disable Homepage Recommendation",
+    "description": "Disable Homepage Recommendation charts",
+    "code": "section[data-testid='home-page'] .main-shelf-shelf:not([aria-label='Recently played'], [aria-label='Your playlists']) { display: none !important; }",
```
kyrie25, line 83: `aria-label` is not the most ideal selector as the label changes for different languages, can you look into a more inclusive one?

packages/marketplace/src/resources/snippets.ts
```diff
@@ -77,4 +77,10 @@ export default [
...
+    "preview": "resources/assets/snippets/disable_recommendation.png",
```
kyrie25, line 84: Use `kebab-case` for file names for uniformity, just a suggestion.

kyrie25 (changes requested):
Hi! Thanks for taking your time on making this PR. I just wanted to point out a few things before we can merge.

packages/marketplace/src/resources/snippets.ts
Tetrax-10: we select this only by `aria-label` because the order changes frequently. The only way to select these without `aria-label` is using JS

kyrie25: It's better that way but not necessarily. Take a look onto the next suggestions and we're good to go.

packages/marketplace/src/resources/snippets.ts
kyrie25: Make the preview image background transparent and delete the cluster also.
image

PR #289, opened 2022-07-29
fix(Grid): only re-render cards on state change

### LINKED ISSUE #106: Don't always re-render every card when a theme is added/removed
**Describe the bug**
Don't always re-render every card when a theme is added/removed. 

**Expected behaviour**
If we can, just re-render the new/old ones.

kyrie25:
Resolves #106 
~~this is part of the reason why cards refused to re-render on search lol~~ as React won't render new components if they have the same key.

### DIFF
--- a/packages/marketplace/src/components/Grid.tsx
+++ b/packages/marketplace/src/components/Grid.tsx
@@ -3,7 +3,7 @@ import semver from "semver";
 import { Option } from "react-dropdown";
 
 import { CardItem, CardType, Config, SchemeIni, Snippet, TabItemConfig } from "../types/marketplace-types";
-import { getLocalStorageDataFromKey, generateSchemesOptions, injectColourScheme } from "../logic/Utils";
+import { getLocalStorageDataFromKey, generateSchemesOptions, injectColourScheme, generateKey } from "../logic/Utils";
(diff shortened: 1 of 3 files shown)


packages/marketplace/src/components/Grid.tsx
```diff
@@ -562,10 +562,11 @@ export default class Grid extends React.Component<
             })
             .map((card) => {
               // Clone the cards and update the prop to trigger re-render
-              const cardElement = React.cloneElement(card, {
+
+              return React.cloneElement(card, {
                 activeThemeKey: this.state.activeThemeKey,
+                key: `${card.props.item.title}-${card.props.item.user}`,
```
theRealPadster, line 568: Do we have or can we make a util function that generates the key based on the card data? Then we don't need to worry about the logic being in two different places (since it's in the card class as well)

theRealPadster (changes requested):
Comment added. Please fix the type annotations thing too.

packages/marketplace/src/components/Grid.tsx
kyrie25: I don't think there is a similar function in the Card component? The key has never been present or generated except for this part.

theRealPadster (approved):
Nice 👌

PR #304, opened 2022-08-25
Feat/export import

CharlieS1103:


CharlieS1103:
Making this pull request as a reminder that we should probably get this done.

Tetrax-10:
you still gonna implement this and add it to marketplace? we have spotify backup.

CharlieS1103:
Yes

theRealPadster:
This isn't done...

CharlieS1103:
@theRealPadster Yeah I did get it working though, essentially all we need to do is make it look nice and it'd work. Although, we may want to transition to loading in the imports through a file rather than the text-area (for the sake of speed.)

theRealPadster:
> @theRealPadster Yeah I did get it working though, essentially all we need to do is make it look nice and it'd work. Although, we may want to transition to loading in the imports through a file rather than the text-area (for the sake of speed.)

I'm taking a look now and though it does copy to clipboard fine, pasting a massive amount of data back in still freezes the client. I think using a file is the only way. Idk how to make it download a file though. My method I was testing with making an `<a>` tag with the `download` attribute set and the `href` set to a blob URI didn't seem to do anything. 

We should be able to use a file input to upload and parse though, since I got that working for the snippet images.

CharlieS1103:
We don't need to download as a file if copying works( probably just impossible), and yeah we can just upload as file

theRealPadster:
> We don't need to download as a file if copying works( probably just impossible), and yeah we can just upload as file

Yeah don't need to, but it would make it easier for people to transfer and back things up if they don't need to paste it into a text editor and save their own.

CharlieS1103:
> > We don't need to download as a file if copying works( probably just impossible), and yeah we can just upload as file
> 
> 
> 
> Yeah don't need to, but it would make it easier for people to transfer and back things up if they don't need to paste it into a text editor and save their own. 

Yeah you're definitely right but if I had to guess, it will be impossible. Might give it a try regardless though.

packages/marketplace/src/components/Modals/ImportExport/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+    // Check if pastedData is valid JSON, if not return an error message and exit
+    try {
+      settings = JSON.parse(pastedData);
+    } catch (e) {
+      Spicetify.showNotification("Invalid JSON");
+      return;
+    }
+    importMarketplace(settings);
+    Spicetify.PopupModal.hide();
+    location.reload();
```
theRealPadster, line 43: Yeah if we're doing a reload here, we don't need it in the import function itself

packages/marketplace/src/components/Modals/ImportExport/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+            placeholder="Copy/paste your settings here"
+            style={{
+              // fontFamily: "'Fira code', 'Fira Mono', monospace'",
+              // fontSize: 12,
+            }}
+          />
+        </div>
+      </div>
+      <>
+        {/* <a href={href} className="download" download="marketplace-export.json">
```
theRealPadster, line 91: Remove if we can't get the download to work

packages/marketplace/src/components/Modals/ImportExport/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+    catch (e) {
+      Spicetify.showNotification("Invalid JSON");
+      return;
+    }
+    importMarketplace(settings);
+
+  };
+
+  return (
+    // TODO: remove "add-snippet" and "customCSS" references
```
theRealPadster, line 69: Remove these references. Probably make class names more generic

packages/marketplace/src/components/Modals/ImportExport/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+import "prismjs/components/prism-json";
+
+import {
+  exportMarketplace,
+  importMarketplace,
+} from "../../../logic/Utils";
+import Button from "../../Button";
+
+const ImportExportModal = () => {
+  const [code, setCode] = React.useState("");
```
theRealPadster, line 13: Name it something more accurate, like settings

packages/marketplace/package.json
```diff
@@ -43,5 +43,9 @@
 		"spicetify-creator": "^1.0.11",
 		"typescript": "^4.8.4"
 	},
-	"private": true
+	"private": true,
+	"dependencies": {
```
theRealPadster, line 47: These should be dev dependencies

PR #307, opened 2022-08-31
Localization (translations)

theRealPadster:
I added [i18next](https://www.i18next.com/) as the library, and added some basic translations for settings modal as a proof of concept. It can detect the language set in Spotify, so we shouldn't need to add an option for it. The translations are stored in different files, so we can have people make PRs and add/update their native language. 

Working on the settings modal so far:
image


theRealPadster:
I think I've got all the strings translated and this is ready to go. Please give it a test and see if I missed anything. I'd also like to note that I left the French "dark" and "light" theme tags in English because I wasn't sure which word referred to a colour theme as opposed to somber/gloomy/etc. @xerta555 What would you use for those?

packages/marketplace/src/resources/locales/en.json
```diff
@@ -0,0 +1,75 @@
...
+      "title": "Settings",
+      "optionsHeading": "Options",
+      "starCountLabel": "Stars count",
+      "tagsLabel": "Tags",
+      "devToolsLabel": "Theme developer tools",
+      "hideInstalledLabel": "Hide installed when browsing",
+      "colourShiftLabel": "Shift colours every minute",
+      "tabsHeading": "Tabs",
+      "resetHeading": "Reset",
+      "resetBtn": "$t(settings.resetHeading)",
```
kyrie25, line 13: Is this necessary when `resetBtn` essentially uses `resetHeading` string?

theRealPadster: In case it ever changes. Technically they serve different purposes. One is a label vs an affirmative action. I think it's best practice, but idk.

kyrie25: It may differ in different languages so I'm not sure whether to remove it so yeah ig we can keep it so.

kyrie25 (approved):
LGTM! Works fine as well.

xerta555:
> @xerta555 What would you use for those?

Not having any programming skills (except shell), all the projects I'm involved in for translations use i18n with sites like Crowdin, GetLocalization, OneSky, Transifex and others.

PR #308, opened 2022-08-31
Add French translation

xerta555:


packages/marketplace/src/resources/locales/fr.json
```diff
@@ -0,0 +1,39 @@
+{
+  "translation": {
+    "settings": {
+      "optionsHeading": "Options",
+      "starCountLabel": "Nombre d\’étoiles",
```
kyrie25, line 5:
```suggestion
      "starCountLabel": "Nombre d’étoiles",
```
Unnecessary & invalid character escape
image

kyrie25 (changes requested):

packages/marketplace/src/resources/locales/fr.json
xerta555: I put the backslash `’` bcz some translations system need this symbols.
So if here it don't wan't the symbol to escape the character, remove it 👍🏻

theRealPadster:
Thanks for this! Can you update the fr-CA one as well? I'm assuming it's basically the same. I just used google translate.

xerta555:
> Thanks for this! Can you update the fr-CA one as well? I'm assuming it's basically the same, so a copy/paste should do it. I just used google translate.

I could do it, but not knowing the subtleties of the Canadian language since i have never practiced it, i don't think i will venture into it.

Other translators can do it when necessary.

theRealPadster:
> I could do it, but not knowing the subtleties of the Canadian language since i have never practiced it, i don't think i will venture into it.
> 
> Other translators can do it when necessary.

Fair point. I think I'll just delete it then, since it was just google translate. It should fall back to using your main "fr" one.

theRealPadster (approved):

xerta555:
> > I could do it, but not knowing the subtleties of the Canadian language since i have never practiced it, i don't think i will venture into it.
> > Other translators can do it when necessary.
> 
> Fair point. I think I'll just delete it then, since it was just google translate. It should fall back to using your main "fr" one.

I think this is the wisest indeed, as i will maybe update my translation to test it in practise.

kyrie25 (approved):

xerta555:
As I thought, the term for this term is missing
image

First question: it could be possible to add it in en file to be able to translate it please ?
Second question: is it possible to compiling the marketplace with an Windows host ?

theRealPadster:
> As I thought, the term for this term is missing image
> 
> First question: it could be possible to add it in en file to be able to translate it please ? Second question: is it possible to compiling the marketplace with an Windows host ?

1. This is not a translatable string. It doesn't work the same way as the others and can not be changed based on language. 
2. Yes, it doesn't matter what OS you're runnning, you just need to have Node.js and Yarn set up in order to run the build script.

PR #316, opened 2022-09-03
feat(snippet): better lyrics style

Tetrax-10:


theRealPadster:
@CharlieS1103 Do you know if these classes are/were ever mapped? If we add this, it'll break next time the classes change.

CharlieS1103:
> @CharlieS1103 Do you know if these classes are/were ever mapped? If we add this, it'll break next time the classes change, like that arrow one.

Lifes a gamble.

Tetrax-10:
> @CharlieS1103 Do you know if these classes are/were ever mapped? If we add this, it'll break next time the classes change, like that arrow one.

If classes change no prob I will regularly check and update every single snippet, themes, extension bro don't worry. I use this in [my theme](https://github.com/Tetrax-10/Nord-Spotify) so if some class change I will update it

kyrie25:
> Do you know if these classes are/were ever mapped? If we add this, it'll break next time the classes change, like that arrow one.

Seems like it has never been, I'll start mapping it from now and we can change the class names there

kyrie25:
> If classes change no prob I will regularly check and update every single snippet, themes, extension bro don't worry. I use this in [my theme](https://github.com/Tetrax-10/Nord-Spotify) so if some class change I will update it

The point is to make snippets easily understood and maintained, also to keep it from bloating up with hashes as class names.

kyrie25:
Mapped `lyrics` classes via spicetify/spicetify-cli#1911
@Tetrax-10 Please change your snippet accordingly so it can take effect next CLI release.

Tetrax-10:
> Mapped `lyrics` classes via [spicetify/spicetify-cli#1911](https://github.com/spicetify/spicetify-cli/pull/1911) @Tetrax-10 Please change your snippet accordingly so it can take effect next CLI release.

updated

theRealPadster (approved):

PR #320, opened 2022-09-06
Added Hide Playing Gif Snippet

notPlancha:
Apparently the Now playing gif can be heavy on the cpu, so this snippet hides it. redd.it/x6pvw0


CharlieS1103 (changes requested):
I'm all good with adding this but maybe remove the part about performance as well as the link to the reddit post because tbh I doubt it.

notPlancha:
Tbh I doubt it too but it was an easy fix that's why I did it
I'll change the text by the morning

theRealPadster:
Can you add like 20-25% whitespace around the edges of the preview? To make it look more consistent with the rest.

notPlancha:
I fixed the ratio to be consistent, hopefully that's what you asked for

theRealPadster:
> I fixed the ratio to be consistent, hopefully that's what you asked for

Not quite what I meant (other icon-only snippets have some transparent padding around the edges), but it looks like it's big enough to fill the snippet preview square, so is probably fine as is.

theRealPadster (approved):

CharlieS1103 (approved):

PR #321, opened 2022-09-06
fix(readme-page): scrollbar not showing up on image load

kyrie25:
weird, but works

Spotify_w4t27w2vJl

### DIFF
--- a/packages/marketplace/src/components/ReadmePage.tsx
+++ b/packages/marketplace/src/components/ReadmePage.tsx
@@ -70,9 +70,12 @@ class ReadmePage extends React.Component<
     // Make the page scrollable
     const main = document.querySelector("#marketplace-readme")?.closest("main");
     if (main) {
-      setTimeout(() => {
+      const callScrollbar = setInterval(() => {
         // TODO: see if it's possible to use some load event or mutation observer to do this
+        main.style.overflowY = "visible";
         main.style.overflowY = "auto";
+        if (!document.querySelector(".os-scrollbar-vertical.os-scrollbar-unusable") || !main)
+          clearInterval(callScrollbar);
       }, 1000);
     }
 


CharlieS1103 (changes requested):
Definitely use a mutation observer here, particularly because the page requires the readme to be fetched

kyrie25:
> Definitely use a mutation observer here, particularly because the page requires the readme to be fetched

Fetching images are not DOM mutations, it doesn't work.
image

kyrie25:
I have also tried `addEventListener("load",` for the `section` part and each element in the Readme also but the engine reports that it is loaded but has actually not finished rendering.

theRealPadster:
How does it take so long to load?

packages/marketplace/src/components/ReadmePage.tsx
```diff
@@ -70,9 +70,12 @@ class ReadmePage extends React.Component<
     // Make the page scrollable
     const main = document.querySelector("#marketplace-readme")?.closest("main");
     if (main) {
-      setTimeout(() => {
+      const callScrollbar = setInterval(() => {
         // TODO: see if it's possible to use some load event or mutation observer to do this
+        main.style.overflowY = "visible";
         main.style.overflowY = "auto";
```
theRealPadster, line 76: Why is it setting this and then resetting it?

kyrie25: So that it works in the way that it is *re-applying* the stylings. Setting it to the same value the second time doesn't change the behavior.

kyrie25:
> How does it take so long to load?

I have zero clue lmfao

packages/marketplace/src/components/ReadmePage.tsx
theRealPadster: This is so messed up 😂

packages/marketplace/src/components/ReadmePage.tsx
```diff
@@ -70,9 +70,12 @@ class ReadmePage extends React.Component<
...
+        if (!document.querySelector(".os-scrollbar-vertical.os-scrollbar-unusable") || !main)
```
theRealPadster, line 77: What's the condition doing here? If there is a scrollbar or if there's no main, stop checking?

kyrie25: When scrollbar is not present, the scrollbar element will have `os-scrollbar-unusable` in class list.
This checks if scrollbar is usable/visible or the user is no longer in the readme page (just to be sure).
Removing the `os-scrollbar-unusable` class name doesn't make it visible tho btw.

theRealPadster: Ah, alright 👍

theRealPadster (approved):

PR #330, opened 2022-09-19
Create codeql-analysis.yml

CharlieS1103:


CharlieS1103:
@theRealPadster This is probably a good thing to add, seems not of our URL inputs are sanitized which we should probably get on tbh.

theRealPadster:
> @theRealPadster This is probably a good thing to add, seems not of our URL inputs are sanitized which we should probably get on tbh.

Yep. I fixed the one issue but haven't gotten around to the others. Feel free to give it a try 😉

CharlieS1103:
> > @theRealPadster This is probably a good thing to add, seems not of our URL inputs are sanitized which we should probably get on tbh.
> 
> Yep. I fixed the one issue but haven't gotten around to the others. Feel free to give it a try 😉

Will do, whenever I find free time lmfao

theRealPadster:
@CharlieS1103 Fixed everything except for the [the one](https://github.com/spicetify/spicetify-marketplace/security/code-scanning/5) dealing with your `getInvalidCSS` function. It seems to think you should change it to use a regex with the `g` option to ensure it replaces all instances in the string. I'm not really sure what the function does, so I think I'll leave that one to you.

theRealPadster (changes requested):
image

PR #339, opened 2022-09-30
feat: update `Spicetify.Config` when loading things in

theRealPadster:


theRealPadster:
I believe this works as intended. Please test it and double check it's working for you. 

My only other todo note is that it feels not quite complete to have it so that it updates the config when installing a no-js theme, but it's unable to reset the config back after removing a no-js theme, since the previous value is lost. 
This would only matter if someone removed a theme and didn't install another on top, since installing one on top would set it again. It's probably not a big deal, just thought I'd mention in case anyone has any thoughts.

packages/marketplace/src/types/spicetify.d.ts
```diff
@@ -1371,9 +1371,13 @@ declare namespace Spicetify {
...
-        const color_scheme: string;
-        const extensions: string[];
+        let current_theme: string;
+        let color_scheme: string;
+        let extensions: string[];
         const custom_apps: string[];
+
+        // These two are just added by Marketplace so we can save and restore them when removing a theme
+        let local_theme: string;
+        let local_color_scheme: string;
```
afonsojramos, line 1381: I think we can change this in `spicetify-cli` as well right? Might as well 🤷

afonsojramos (approved):

packages/marketplace/src/types/spicetify.d.ts
kyrie25: We could but it's meant to be a copy, not a way to modify `config-xpui` from within the app. The only edge case where we would only need to do this would be Marketplace tbh.

theRealPadster: Does it make sense to save `local_theme` in cli? Or not really since `current_theme` is meant to be local?

kyrie25: The latter, yeah. Unless we do want to make Marketplace a part of CLI, there aren't any reason to have both properties have the same value since currently it takes stuff from `config-xpui`

PR #347, opened 2022-10-09
Rotating Cover Art

ItsRemix:
# CSS Snippet to add rotating cover art to spotify

cover


packages/marketplace/src/resources/snippets.ts
```diff
@@ -155,4 +155,10 @@ export default [
     "code": `.main-collectionLinkButton-collectionLinkButton[href="/collection/tracks"], .main-collectionLinkButton-collectionLinkButton[href="/collection/episodes"] {display: none;}`,
     "preview": "resources/assets/snippets/remove-ep-likes.png",
   },
+  {
+    "title": "Vinyl Cover Art",
```
theRealPadster, line 159: Maybe change the name to "Rotating Cover Art", since that's what's in the preview image?

theRealPadster (changes requested):
Tested it myself and seems good. 👍  Just made a suggestion for naming. Could you also fix the missing trailing comma linting error?

packages/marketplace/src/resources/snippets.ts
ItsRemix: Ye changed it! I also fixed linting error. Thanks! <3

theRealPadster: Thanks :) Still missing a comma at the end of the preview line

ItsRemix: Fixed!

theRealPadster: 👌

theRealPadster (approved):

ItsRemix:
btw how much it takes it ot get to app

theRealPadster:
I think we fetch them via JS, so should be there if you restart Spotify?

ItsRemix:
i dont have it

theRealPadster:
Oh, looks like it does use the ones locally. In that case you'll need to update your marketplace to the latest build from the dist branch. 

https://github.com/spicetify/spicetify-marketplace/blob/main/packages/marketplace/src/logic/FetchRemotes.ts

PR #348, opened 2022-10-13
Fix "Hide Upgrade Button"

whyredow:
image


kyrie25:
Upgrade button class already mapped on CLI. Run `spicetify restore backup apply`

PR #353, opened 2022-10-29
adding custom Background

byQuexo:
Purple Darth Vadder Background

### DIFF
--- a/packages/marketplace/src/resources/snippets.ts
+++ b/packages/marketplace/src/resources/snippets.ts
@@ -164,6 +164,6 @@ export default [
   {
     "title": "Custom Background",
     "description": "Adds a custom background to your Spicetify, you can Change the Image with replacing the url in the body.",
-    "code": "body { background-image: url('https://imgur.com/gallery/848rL'); }",
+    "code": "body { background-image: url('https://i.imgur.com/ztoFEP5.jpeg'); }",
   },
 ];
--- a/packages/marketplace/src/resources/snippets.ts
+++ b/packages/marketplace/src/resources/snippets.ts
@@ -161,4 +161,9 @@ export default [
     "code": `@keyframes rotating { from { transform: rotate(0deg); } to { transform: rotate(360deg); }} .cover-art { animation: rotating 2s linear infinite; clip-path: circle(50% at 50% 50%);}`,
     "preview": "resources/assets/snippets/rotating-coverart",
   },
+  {
+    "title": "Custom Background",
+    "description": "Adds a custom background to your Spicetify, you can Change the Image with replacing the url in the body.",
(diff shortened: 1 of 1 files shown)


xerta555:
Why don't adding an option into the Marketplace options with a toggle to enable/disable the option itself, and a second one to let or change the default URL you purpose ?
It would be more flexible..

CharlieS1103:
> Why don't adding an option into the Marketplace options with a toggle to enable/disable the option itself, and a second one to let or change the default URL you purpose ?
> 
> It would be more flexible..

Bit confused on what you mean by the first part

xerta555:
> > Why don't adding an option into the Marketplace options with a toggle to enable/disable the option itself, and a second one to let or change the default URL you purpose ?
> > It would be more flexible..
> 
> Bit confused on what you mean by the first part

In simple therms: why just don't add an enable/disable option to apply a custom background img ?
A first option to toggle the function and a second one to apply a specific one with an URL.

kyrie25:
> > > Why don't adding an option into the Marketplace options with a toggle to enable/disable the option itself, and a second one to let or change the default URL you purpose ?
> 
> > > It would be more flexible..
> 
> > 
> 
> > Bit confused on what you mean by the first part
> 
> 
> 
> In simple therms: why just don't add an enable/disable option to apply a custom background img ?
> 
> A first option to toggle the function and a second one to apply a specific one with an URL.

That would not be possible using pure CSS/as a snippet, which is what this PR tries to do.

CharlieS1103:
I had an idea to make this viable, let snippets use string and int variables and expand their features a little to allow users to input these

kyrie25:
> I had an idea to make this viable, let snippets use string and int variables and expand their features a little to allow users to input these 

Lol if so just let them use CSS body variables. You can set them using snippets already.

theRealPadster:
Do we let users edit the built-in snippets? Then they could just change the image url

CharlieS1103:
> Do we let users edit the built-in snippets? Then they could just change the image url

Pretty sure we do, but I think basically wrapping css variables in a ui is probably more useable since a lot of users wouldn't want to edit the text

Satanarious:
The [Spicetify Galaxy](https://github.com/harbassan/spicetify-galaxy) theme lets you use custom backgrounds for Home Screen(This is applied to all other pages other than the forth mentioned one) and Playlist Pages. You can maybe request him to add an option for the marketplace. He's quick with the replies and added a feature I requested in a day or two.

theRealPadster:
We could also just add it as an option in the marketplace settings, and use a variable. 

e.g. in the settings modal, have a text entry for adding an image url, and that's saved as `--marketplace-body-background`. Then when that's changed, we set it on body. Or is it possible to have CSS be conditional if a variable exists or not?

PR #354, opened 2022-10-30
Cleanup import-export feature

CharlieS1103:
@theRealPadster Should account for everything referenced in your review

### DIFF
--- a/packages/marketplace/src/components/Modals/ImportExport/index.tsx
+++ b/packages/marketplace/src/components/Modals/ImportExport/index.tsx
@@ -24,7 +24,7 @@ const ImportExportModal = () => {
   };
 
   const importSettings = () => {
-    const pastedData : string = importText;
+    const pastedData: string = importText;
     let settings : JSON;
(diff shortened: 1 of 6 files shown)


packages/marketplace/src/components/Modals/ImportExport/index.tsx
```diff
@@ -24,7 +24,7 @@ const ImportExportModal = () => {
   };
 
   const importSettings = () => {
-    const pastedData : string = code;
+    const pastedData : string = importText;
```
ririxidev, line 27:
```suggestion
    const pastedData: string = importText;
```

ririxidev (changes requested):
🙈

packages/marketplace/src/components/Modals/ImportExport/index.tsx
CharlieS1103: we don't talk about it.

ririxidev (approved):

PR #355, opened 2022-10-30
Bump package.json

CharlieS1103:


CharlieS1103:
@theRealPadster Should we bump it before or after creating the release

kyrie25 (commented):
Bump it inside the Marketplace package also.

theRealPadster:
> @theRealPadster Should we bump it before or after creating the release

Before. I botched the release tho 🙈 we need to delete the tag and redo it

PR #369, opened 2022-11-04
i18n: Add translation support for import/export functionality

theRealPadster:
Enable translation support for import/export modal. Also renames it to "backup modal" and tweaks some of the copy. 

image

image


CharlieS1103 (approved):

theRealPadster:
I want to rename the modal to be consistent with the settings menu. Call it `BackupModal` instead of `ImportExportModal`?

CharlieS1103 (changes requested):
We should probably make it work first.

theRealPadster:
> We should probably make it work first.

Fixed

xerta555:
My pr: https://github.com/spicetify/spicetify-marketplace/pull/370

kyrie25 (approved):

PR #370, opened 2022-11-04
Update French translation

xerta555:


packages/marketplace/src/resources/locales/fr.json
```diff
@@ -42,6 +45,17 @@
       "reloadNow": "Recharger maintenant",
       "reloadLater": "Recharger plus tard"
     },
+    "backupModal": {
+      "title": "Importer/Exporter",
```
theRealPadster, line 49: The title in English is "Back up/Restore"

theRealPadster (changes requested):

packages/marketplace/src/resources/locales/fr.json
xerta555: Strange, i see this in the en file
image

kyrie25: > Strange, i see this in the en file image

That would be this, which is also `Back up/Restore`:
image

kyrie25: You can reuse strings btw, in case you haven't noticed.

xerta555: Ok done.

theRealPadster (approved):

theRealPadster:
Thanks!

PR #372, opened 2022-11-08
feat: add new snippets

Plueres:
added a custom snippet to align the playlists and folders on the left sidebar to all the other categories in the sidebar.


Plueres:
The first part of the code can be replaced with anything shorter, but for some reason when I remove the line, it reverts the last few lines.

packages/marketplace/src/resources/snippets.ts
```diff
@@ -167,4 +167,10 @@ export default [
     "code": ".collection-collectionEntityHeroCard-likedSongs{ display: none; }",
     "preview": "resources/assets/snippets/hide-likedSongs-card.png",
   },
+  {
+    "title": "Fix playlist and folder position",
+    "description": "Replaces the playlists section in the left side bar to align with all the other category icons for the new UI. Screenshot includes "Playlist icons" snippet",
```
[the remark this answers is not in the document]
Plueres: Alright, will do!

Plueres:
made the changes in the latest commit

theRealPadster:
Please include the preview image file itself

Plueres:
> Please include the preview image file itself

alright, inside the same folder as all the others?

Plueres:
I hope it is fixed now

packages/marketplace/src/resources/snippets.ts
```diff
@@ -167,4 +167,10 @@ export default [
     "code": ".collection-collectionEntityHeroCard-likedSongs{ display: none; }",
     "preview": "resources/assets/snippets/hide-likedSongs-card.png",
   },
+  {
+    "title": "Fix playlist and folder position",
+    "description": "Replaces the playlists section in the left side bar to align with all the other category icons for the new UI. Screenshot includes 'Playlist icons' snippet",
+    "code": "{} .main-rootlist-rootlistItem { height: 40px; } .gJFKvJ { font-weight: 700!important; } .playlist-item__img { margin-right: 16px!important; }",
```
theRealPadster, line 173: Is there a reason why there's an empty {} at the beginning? 
I'm also wondering if we need to map some CSS selector, since that .gJFKvJ is likely to break down the road.

theRealPadster: @kyrie25 ^

Plueres: The {} has to be there for some unknown reason, otherwise it won't output the expected result. I've noticed that it can't be a different symbol, but it can be like I did in the first commit.

theRealPadster: It acts the exact same for me with and without the beginning `{}`

kyrie25: > @kyrie25 ^

It serves no purpose afaik.

kyrie25: Also hashes as selectors aren't recommended, use relative selectors instead, (`>`, `+`, `~`, `nth-child`, etc.)

Plueres: > It acts the exact same for me with and without the beginning `{}`

For me it removes the height adjustment if I don't have anything in front/in the place of the `{}`. Should I just commit it without the brackets?

Plueres: > Also hashes as selectors aren't recommended, use relative selectors instead, (`>`, `+`, `~`, `nth-child`, etc.)
> 
> > I'm also wondering if we need to map some CSS selector, since that .gJFKvJ is likely to break down the road.
> 
> 6-character classes are dynamically generated and cannot be mapped unfortunately.

Is there anything else I should adjust

kyrie25: > For me it removes the height adjustment if I don't have anything in front/in the place of the `{}`. Should I just commit it without the brackets?

Yes.

> Is there anything else I should adjust

LGTM.

kyrie25: > > LGTM. As soon as you address the selector it should be fine.
> > Also hashes as selectors aren't recommended, use relative selectors instead, (>, +, ~, nth-child, etc.)
> 
> Could you maybe be more specific in which one you mean? Are the classes not correct? Edit: do you mean from the !importants?

Selector is what the CSS ruleset applies on, in this case I'm addressing the one Padster mentioned above - `.gJFKvJ`.
If possible, use relative selectors instead, eg. `.main-<section>-<component> > span`

Plueres: Alright, done! I hope the css is correct now.

theRealPadster: That selector looks better to me. Haven't tested but I trust it does the same thing?

Plueres: I think the previous selector was only for a specific folder... now it selects all folders **and** playlists, and now it actually works. Sadly I still had to keep the !importants

kyrie25:
> Yo @kyrie25, it says it's block because you still have requested changes. Are we good now?

All good.

kyrie25 (approved):

PR #394, opened 2022-12-03
feat: Album-Art Based Color Shifting

CharlieS1103:
Going to edit the title before we merge, just thought it'd be funny.

This code isn't as clean as it can be of course, which is what I assume the review process will help with, however it does work.

New libraries/ api's used
When the song changes, it will get the main color of the album art using **node vibrant**,
it will then generate a scheme using **thecolorapi**
and the final library used is **chroma**, in order to sort the scheme the user actively has installed by the amount of contrast between each color 

To elaborate on some things, the reason I sort the colorMap based on the relative distance between the colors is so I can get it to approximately match the order that the schemes colorapi provides are ordered in, meaning the text and the main colors should always contrast each other rather nicely. This does cause all the schemes to have a dark background with a light text so I might allow the users to change the mode colorapi generates the schemes or change the way I order them (use saturation, lumination, or some other value).

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,5 +15,9 @@
 	"packageManager": "yarn@3.2.1",
 	"devDependencies": {
 		"husky": "^8.0.2"
+	},
+	"dependencies": {
+		"@types/chroma-js": "^2.1.4",
+		"chroma-js": "^2.4.2"
 	}
 }
--- a/package.json
+++ b/package.json
@@ -15,9 +15,5 @@
 	"packageManager": "yarn@3.2.1",
 	"devDependencies": {
 		"husky": "^8.0.2"
-	},
-	"dependencies": {
-		"@types/chroma-js": "^2.1.4",
-		"chroma-js": "^2.4.2"
 	}
 }
--- a/packages/marketplace/package.json
+++ b/packages/marketplace/package.json
@@ -45,5 +45,8 @@
 		"spicetify-creator": "^1.0.11",
 		"typescript": "^4.9.3"
 	},
-	"private": true
+	"private": true,
+	"dependencies": {
+		"node-vibrant": "3.1.4"
+	}
 }
--- a/packages/marketplace/package.json
+++ b/packages/marketplace/package.json
@@ -24,6 +24,7 @@
 		"node": ">=16"
 	},
 	"devDependencies": {
+		"@types/chroma-js": "^2.1.4",
 		"@types/filesystem": "^0.0.32",
 		"@types/react": "^17.0.2",
 		"@types/react-dom": "^17.0.2",
@@ -31,11 +32,13 @@
 		"@types/wicg-file-system-access": "^2020.9.5",
 		"@typescript-eslint/eslint-plugin": "^5.45.0",
 		"@typescript-eslint/parser": "^5.45.0",
+		"chroma-js": "^2.4.2",
 		"copyfiles": "^2.4.1",
 		"eslint": "^8.28.0",
 		"eslint-plugin-react": "^7.31.11",
 		"i18next": "^22.0.6",
 		"i18next-browser-languagedetector": "^7.0.1",
+		"node-vibrant": "3.1.4",
 		"prismjs": "^1.29.0",
 		"react-dropdown": "^1.11.0",
 		"react-i18next": "^12.0.0",
@@ -45,8 +48,5 @@
 		"spicetify-creator": "^1.0.11",
 		"typescript": "^4.9.3"
 	},
-	"private": true,
-	"dependencies": {
-		"node-vibrant": "3.1.4"
-	}
+	"private": true
 }
--- a/packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
+++ b/packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
@@ -9,11 +9,11 @@ const ConfigRow = (props: {
   modalConfig: Config;
   clickable?: boolean;
   updateConfig: (CONFIG: Config) => void;
-  type: string | null;
-  options: string[] | null;
+  type?: string;
+  options?: string[];
 }) => {
   const type = props.type;
-  const componentId = type == "dropdown" ? "dropdown:" +props.storageKey : "toggle:" + props.storageKey;
+  const componentId = type === "dropdown" ? "dropdown:" + props.storageKey : "toggle:" + props.storageKey;
   const enabled = !!props.modalConfig.visual[props.storageKey];
 
   const settingsToggleChange = (e) => {
--- a/packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
+++ b/packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
@@ -13,7 +13,9 @@ const ConfigRow = (props: {
   options?: string[];
 }) => {
   const type = props.type;
-  const componentId = type === "dropdown" ? "dropdown:" + props.storageKey : "toggle:" + props.storageKey;
+  const componentId = (type === "dropdown")
+    ? "dropdown:" + props.storageKey
+    : "toggle:" + props.storageKey;
   const enabled = !!props.modalConfig.visual[props.storageKey];
 
   const settingsToggleChange = (e) => {
--- a/packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
+++ b/packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
@@ -35,10 +35,9 @@ const ConfigRow = (props: {
     props.updateConfig(props.modalConfig);
   };
 
-  if (type == "dropdown" && props.options != null) {
+  if (type === "dropdown" && props.options) {
     return (
       <SortBox
-
         sortBoxOptions={props.options.map((option) => {
           return {
             key: option,
--- a/packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
+++ b/packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
@@ -2,15 +2,18 @@ import React from "react";
 import { Config } from "../../../types/marketplace-types";
 
 import Toggle from "../../Toggle";
-
+import SortBox from "../../Sortbox";
 const ConfigRow = (props: {
   name: string;
   storageKey: string;
   modalConfig: Config;
   clickable?: boolean;
   updateConfig: (CONFIG: Config) => void;
+  type: string | null;
+  options: string[] | null;
 }) => {
-  const toggleId = `toggle:${props.storageKey}`;
+  const type = props.type;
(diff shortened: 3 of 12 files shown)


packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
```diff
@@ -2,15 +2,18 @@ import React from "react";
...
   storageKey: string;
   modalConfig: Config;
   clickable?: boolean;
   updateConfig: (CONFIG: Config) => void;
+  type: string | null;
+  options: string[] | null;
 }) => {
-  const toggleId = `toggle:${props.storageKey}`;
+  const type = props.type;
+  const componentId = type == "dropdown" ? "dropdown:" +props.storageKey : "toggle:" + props.storageKey;
```
kyrie25, line 16: Good practice, also strict equal cuz loose equal is just funky
```suggestion
  const componentId = type === "dropdown" ? "dropdown:" + props.storageKey : "toggle:" + props.storageKey;
```

packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
```diff
@@ -24,10 +27,35 @@ const ConfigRow = (props: {
...
+    props.modalConfig.visual[storageKey] = state;
+    localStorage.setItem(`marketplace:${storageKey}`, String(state));
+    props.updateConfig(props.modalConfig);
+  };
+
+  if (type == "dropdown" && props.options != null) {
+    return (
+      <SortBox
+
+        sortBoxOptions={props.options.map((option) => {
```
kyrie25, line 42:
```suggestion
  if (type === "dropdown" && props.options) {
    return (
      <SortBox
        sortBoxOptions={props.options.map((option) => {
```

packages/marketplace/src/components/Modals/Settings/index.tsx
```diff
@@ -43,11 +43,15 @@ const SettingsModal = ({ CONFIG, updateAppConfig } : Props) => {
...
-      <ConfigRow name={t("settings.colourShiftLabel")} storageKey='colorShift' modalConfig={modalConfig} updateConfig={updateConfig} />
+      <ConfigRow name={t("settings.starCountLabel")} storageKey='stars' modalConfig={modalConfig} updateConfig={updateConfig} type={null} options={null} />
+      <ConfigRow name={t("settings.tagsLabel")} storageKey='tags' modalConfig={modalConfig} updateConfig={updateConfig} type={null} options={null} />
+      <ConfigRow name={t("settings.devToolsLabel")} storageKey='themeDevTools' modalConfig={modalConfig} updateConfig={updateConfig} type={null} options={null} />
+      <ConfigRow name={t("settings.hideInstalledLabel")} storageKey='hideInstalled' modalConfig={modalConfig} updateConfig={updateConfig} type={null} options={null} />
+      <ConfigRow name={t("settings.colourShiftLabel")} storageKey='colorShift' modalConfig={modalConfig} updateConfig={updateConfig} type={null} options={null} />
+      <ConfigRow name={t("settings.albumArtBasedColors")} storageKey='albumArtBasedColors' modalConfig={modalConfig} updateConfig={updateConfig} type={null} options={null} />
```
kyrie25, line 51: Revert these

packages/marketplace/src/constants.ts
```diff
@@ -12,7 +12,10 @@ export const LOCALSTORAGE_KEYS = {
   sortBy: "marketplace:sort-by",
   // Theme installed store the localsorage key of the theme (e.g. marketplace:installed:NYRI4/Comfy-spicetify/user.css)
   themeInstalled: "marketplace:theme-installed",
+  albumArtBasedColor: "marketplace:albumArtBasedColors",
+  albumArtBasedColorMode: "marketplace:albumArtBasedColorsMode",
   colorShift: "marketplace:colorShift",
+
 };
```
kyrie25, line 19:
```suggestion
  colorShift: "marketplace:colorShift",
};
```

packages/marketplace/src/components/Modals/Settings/ConfigRow.tsx
```diff
@@ -2,15 +2,18 @@ import React from "react";
...
-
+import SortBox from "../../Sortbox";
 const ConfigRow = (props: {
   name: string;
   storageKey: string;
   modalConfig: Config;
   clickable?: boolean;
   updateConfig: (CONFIG: Config) => void;
+  type: string | null;
+  options: string[] | null;
```
kyrie25, line 13: Pass parameters as optional so you don't have to do all that funky thing below lol
```suggestion
  type?: string;
  options?: string[];
```

packages/marketplace/src/extensions/extension.tsx
```diff
@@ -91,8 +92,10 @@ import {
 
       // Add to Spicetify.Config
       Spicetify.Config.color_scheme = themeManifest.activeScheme;
-
-      if (localStorage.getItem(LOCALSTORAGE_KEYS.colorShift) === "true") {
+      if (localStorage.getItem(LOCALSTORAGE_KEYS.albumArtBasedColor) === "true") {
+        initAlbumArtBasedColor(activeScheme);
+      }
+      else if (localStorage.getItem(LOCALSTORAGE_KEYS.colorShift) === "true") {
```
kyrie25, line 98:
```suggestion
      } else if (localStorage.getItem(LOCALSTORAGE_KEYS.colorShift) === "true") {
```

packages/marketplace/src/logic/Utils.ts
```diff
@@ -19,6 +19,7 @@ export const getLocalStorageDataFromKey = (key: string, fallback?: unknown) => {
       return data;
     }
   } else {
+    console.log(fallback);
```
kyrie25, line 22: You shouldn't need to `console.log` this if we were going to merge this to `main`?

packages/marketplace/src/logic/Utils.ts
```diff
@@ -317,6 +318,109 @@ export const initColorShiftLoop = (schemes: SchemeIni) => {
...
+      console.error(err);
+      return;
+    }
+    return palette;
+  });
+  if (swatches.Vibrant) {
+    // remove the # from the hex
+    return swatches.Vibrant.hex.substring(1);
+
+  }
```
kyrie25, line 333: Why are there do many random newlines lol
```suggestion
    return swatches.Vibrant.hex.substring(1);
  }
```

packages/marketplace/src/logic/Utils.ts
```diff
@@ -317,6 +318,109 @@ export const initColorShiftLoop = (schemes: SchemeIni) => {
...
+  return "null";
+
+};
+
+export const generateColorPalette = async (mainColor : string, numColors : number) => {
+  // Generate a palette from https://www.thecolorapi.com/id?hex=0047AB&rgb=0,71,171&hsl=215,100%,34%&cmyk=100,58,0,33&format=html
+  const mode = getLocalStorageDataFromKey("marketplace:albumArtBasedColorsMode");
+  // Add a hyphen before any uppercase characters
+  const modeStr = mode.replace(/([A-Z])/g, "-$1").toLowerCase();
+  const palette = await Spicetify.CosmosAsync.get(`https://www.thecolorapi.com/scheme?hex=${mainColor}&mode=${modeStr}&count=${numColors}`);
```
kyrie25, line 343: `CosmosAsync` includes a bunch of Spotify authorization stuff onto its header, couldn't this be done normally via `fetch`?

packages/marketplace/src/logic/Utils.ts
```diff
@@ -317,6 +318,109 @@ export const initColorShiftLoop = (schemes: SchemeIni) => {
...
+
+export const initAlbumArtBasedColor = (scheme: ColourScheme) => {
+  const style = document.createElement("style");
+  style.className = "colorShift-style";
+  style.innerHTML = `* {
+    transition-duration: 400ms;
+  }
+  main-type-bass {
+    transition-duration: unset !important;
+  }`;
```
kyrie25, line 369: Lol what
```suggestion
  .main-type-bass {
    transition-duration: unset !important;
  }`;
```

packages/marketplace/src/logic/Utils.ts
```diff
@@ -378,7 +482,7 @@ export const getParamsFromGithubRaw = (url: string) => {
 
 export function addToSessionStorage(items, key?) {
   if (!items) return;
-  items.forEach(item => {
+  items.forEach((item: any) => {
```
kyrie25, line 485: Just revert this to avoid the warning lol

packages/marketplace/src/types/marketplace-types.d.ts
```diff
@@ -116,6 +117,8 @@ export type VisualConfig = {
   hideInstalled: boolean;
   colorShift: boolean;
   themeDevTools: boolean;
+  albumArtBasedColors: boolean;
+  albumArtBasedColorsMode: "monochromeLight" | "monochromeDark" | "quad" | "triad" | "analogic"|"analogicComplement";
```
kyrie25, line 121:
```suggestion
  albumArtBasedColorsMode: "monochromeLight" | "monochromeDark" | "quad" | "triad" | "analogic" | "analogicComplement";
```

kyrie25 (changes requested):

packages/marketplace/src/logic/Utils.ts
CharlieS1103: Forgot to remove this lol

packages/marketplace/src/logic/Utils.ts
CharlieS1103: Spotify got funky with it for whatever reason, it worked sometimes but not others and it didn't appear to be ratelimiting

kyrie25: We used `fetch` everywhere else in Marketplace, not exclusively this. Surely it could not have been so.

CharlieS1103: I could be misremembering with an error I experienced when attempting to use no api's, I don't think i am, but i will try tomorrow nontheless.

packages/marketplace/package.json
```diff
@@ -45,5 +45,8 @@
 		"spicetify-creator": "^1.0.11",
 		"typescript": "^4.9.3"
 	},
-	"private": true
+	"private": true,
+	"dependencies": {
+		"node-vibrant": "3.1.4"
```
theRealPadster, line 50: These should be dev-deps

package.json
```diff
@@ -15,5 +15,9 @@
 	"packageManager": "yarn@3.2.1",
 	"devDependencies": {
 		"husky": "^8.0.2"
+	},
+	"dependencies": {
+		"@types/chroma-js": "^2.1.4",
+		"chroma-js": "^2.4.2"
```
theRealPadster, line 21: These should be dev-deps

PR #396, opened 2022-12-04
Remove extra stuff and workspaces

theRealPadster:
WIP


.github/CODEOWNERS
```diff
@@ -1,3 +1,3 @@
 * @spicetify/marketplace
```
theRealPadster, line 1: I'm not sure what this line does tbh. Does it need to be adjusted as well or no?

package.json
```diff
@@ -1,19 +1,50 @@
...
+		"lint:ci": "eslint src",
+		"type-check": "tsc --noEmit",
+		"watch": "spicetify-creator --watch",
+		"prepare": "husky install"
+	},
+	"engines": {
+		"node": ">=16"
 	},
-	"private": true,
-	"packageManager": "yarn@3.2.1",
```
theRealPadster, line 15: Oh, do we need this still?

package.json
```diff
@@ -1,19 +1,50 @@
...
 	"devDependencies": {
-		"husky": "^8.0.2"
```
theRealPadster, line 17: And this? I just moved the package.json from marketplace to the main one and overwrote it.

.github/CODEOWNERS
kyrie25: Nah, that just means every file in the repository is owned by the `@spicetify/marketplace` team

package.json
kyrie25: Those can stay I think, using a new version of Yarn is always nice

package.json
kyrie25: Husky is for linting before committing, that should stay imo.

theRealPadster:
Does anything else need tweaking?

CharlieS1103 (approved):

PR #413, opened 2023-01-16
feat: Add TooltipWrapper to icon buttons

theRealPadster:
This adds the tooltipwrapper to the theme dev tools and settings buttons. For some reason when I tried to add it to the buttons on the cards, the tooltip doesn't show up.


theRealPadster:
I'll mark it as ready and we can merge, or someone can try and figure out why it doesn't seem to work when I wrap the Card button with it.

kyrie25 (approved):

PR #419, opened 2023-01-31
fix: marketplace item injection location

Tetrax-10:
Injects marketplace items in body like CLI instead of head


src/logic/Utils.ts
```diff
@@ -242,8 +242,8 @@ export const injectColourScheme = (scheme: ColourScheme | null) => {
     let injectStr = ":root {";
     const themeIniKeys = Object.keys(scheme);
     themeIniKeys.forEach((key) => {
-      injectStr += `--spice-${key}: #${scheme[key]} !important;`;
-      injectStr += `--spice-rgb-${key}: ${hexToRGB(scheme[key])} !important;`;
```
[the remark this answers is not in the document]
kyrie25: #287

rxri:
I don't really think that this stuff should be in `<body>` tag. `<style>` tag must be in head no matter of what

rxri:
I don't really think that this stuff should be in `<body>` tag. `<style>` tag must be in head regardless

src/logic/Utils.ts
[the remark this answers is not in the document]
Tetrax-10: Actually we dont need that `!important` because this PR actually removes the `color.css` like `user.css` used to get removed when installing a theme from marketplace.
image

Marketplace devs just forgot to remove `color.css` while installing a theme from marketplace as they actually wrote code to re-inject `color.css` again when a theme is uninstalled from marketplace.
image

#287 @kyrie25 just fixed this problem by adding a `!important` instead of removing the old `color.css`

Tetrax-10:
@rxri actually you are correct CSS needs to be put inside `</head>`, but when they are the original css for a element.

When it comes to overwritting rules (theming / css injection) and snippets we need to inject them to the body as the CSS takes the last value it reads. where head is read first and then body is read at last.

Me and Tominator has faced this overwritting rules by spotify rules for `Nord Spotify` and `Comfy` respectively, As some rules just appends to the dom dynamically by spotify, those rules just overwrites any previous rules defined (usually themes) some times even `!important` once (If spotify rules had `!important`). To fix this I made Nord Spotify to inject the CSS to the body with JS help and It solved this issue.

Also we only face this rule overwritting issue when theme installed through marketplace not the CLI one. As CLI injects CSS to the body and it works as intended

If we want the Theme to work excatly like they are installed through CLI then we need to inject css to the body.

As theme devs install theme locally and test them and finally publish them to marketplace. But when installed through marketpalce to check how it looks some styles wont look as intended as they are overwritten by the styles that are injected to the body by spotify. Like the below example

image

</br>

## when css injected to head (ignore that settings pannel its because of custom apps just forget that)
image

## when css injected to body (intended look)
image

rxri:
Well because order matters. However, it still should be put in `<head>`. A so called "feature" browsers implemented, is allowed, but it's not really recommended

> As CLI injects CSS to the body and it works as intended

it is correct, however it's injected by `link` that is "allowed" in body. cli doesn't use `<style>` tag to inject css
image

src/logic/Utils.ts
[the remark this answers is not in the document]
Tetrax-10: Ok i see that you removed those lines and devs didnt forget that. I dont know why you removed it. But removing `color.css` should remove all the `--spice-something` variable

### U can see there is no `:root` that contains `--spice-something` variables
https://user-images.githubusercontent.com/75513645/215853558-d9371c90-f0e5-41be-83d2-42fae83ffb32.mp4

Other that `color.css` where do you find `--spice-something` variables kyrie?

Tetrax-10:
## Its fine to have style tags in body

image

image


why do we need to worry about the orders? where the theme is bugged when CSS is injected to the head. So you want theme devs to publish their code and then check page by page, element by element if its working perfectly in marketplace version?

Nord Spotify just puts stlyle tag in body for ton of Snippets, there have not been a single problem since.

image

theRealPadster:
Can you post the links for the screenshots? I'd like to read more, that's interesting... [MDN says](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style#technical_summary) it needs `scoped` if it's in the body. I think you're right about us probably forgetting to remove the color.ini stylesheet, unless we left it there on purpose in case a theme doesn't ship its own colour schemes.

rxri:
After bit of testing, I can't find any differences between body and head. Nord theme is all the same in both places. 
image 
image
image I only noticed that some styles don't apply when `!important` is in some style, but that doesn't change when I inject it in body

Tetrax-10:
[link](https://stackoverflow.com/questions/2830296/using-style-tags-in-the-body-with-other-html) you asked for

### you can see `color.ini` are compulsary and `user.css` is optional
image

### And if there is no `color.ini` file then spicetify just creates a default `color.css` file with the spotify's color value and injects it to the DOM
image

Tetrax-10:
@rxri Of cource Nord Spotify will work fine because my `user.css` and `color.css` are empty and I Inject them remotely with JS
image
image

Those screenshot are from september 2022 from discord chat. I wanted to fix it from them but didnt have time so i did today

rxri:
Like I mentioned, I disabled JavaScript injection and injected `src/nord.css` with marketplace instead

Tetrax-10:
i dont get it wdym?

you can see no diff in marketplace Nord spotify and local Nord Spotify as they all gettings injected to body. no matter where the get installed from.

rxri:
I modified nord theme manifest inside localStorage to remove included javascript and changed css url to `src/nord.css`. Marketplace injected this css to DOM. Later, I changed code to inject it to `<body>` and didn't see any differences

rxri:
Never mind. Let's pretend this conversation never happened. I worked too much with classes that I forgot how specificity works in css

Tetrax-10:
That fine 🤗, this was really a health conversation as you really check the code is worth to merge. Appr it 👍

Tetrax-10:
@theRealPadster  You where right. I didnt think about marketplace theme's `color.ini`.

Themes like BurntSienna have less variables so If we remove `color.css` then some `--spice-variables` will go missing. So I reverted that.
image

@kyrie25 We still don't need `!important` as we inject style to `body` marketplace will inject the color scheme only after local `color.css` as they are hard coded in `index.html`
image

Now marketplace will 100% overwrite local `color.css`. Before you used `!important` as they are injected in head (less priority). Just test this build yourself you will get it
image

theRealPadster:
> [link](https://stackoverflow.com/questions/2830296/using-style-tags-in-the-body-with-other-html) you asked for

I took a look at the post and relevant links, and it looks like that last line ("In the body, where flow content is expected") is no longer [in the spec](https://html.spec.whatwg.org/multipage/semantics.html#the-style-element), and `scoped` css [is not supported](https://caniuse.com/style-scoped) on any browsers anymore

You said Spotify itself adds style tags in the body? The only ones I see are in `<head>` or from Nord-Spotify (or my hide podcasts extension apparently 🙃 )

Tetrax-10:
Yeah scoped style tags are gone that's the matter so it's safe to put style tags In body. (That's what I posted in that 2nd ss)

Also i never said Spotify injects style tags to body. I just said they include styles dynamically. You can open network tab and see sometimes css files are imported dynamically.

Yesterday I clearly explained everything to @rxri and she knows why marketplace styled need to be injected to body.

@rxri can you explain him?

src/extensions/extension.tsx
```diff
@@ -35,10 +35,10 @@ import {
   // https://github.com/satya164/react-simple-code-editor/issues/86
   const reactSimpleCodeEditorFix = document.createElement("script");
   reactSimpleCodeEditorFix.innerHTML = "const global = globalThis;";
-  document.head.appendChild(reactSimpleCodeEditorFix);
+  document.body.appendChild(reactSimpleCodeEditorFix);
 
   // Show message on start.
-  console.log(`Initializing Spicetify Marketplace v${MARKETPLACE_VERSION}`);
+  // console.log(`Initializing Spicetify Marketplace v${MARKETPLACE_VERSION}`);
```
theRealPadster, line 41: Can we leave this one in? We added it so it's easy to see if someone has marketplace installed properly.

src/logic/Utils.ts
```diff
@@ -242,19 +241,12 @@ export const injectColourScheme = (scheme: ColourScheme | null) => {
...
+      injectStr += `--spice-${key}: #${scheme[key]};`;
+      injectStr += `--spice-rgb-${key}: ${hexToRGB(scheme[key])};`;
     });
     injectStr += "}";
     schemeTag.innerHTML = injectStr;
-    document.head.appendChild(schemeTag);
```
theRealPadster, line 250: Why is this stuff removed again?

src/app.tsx
```diff
@@ -100,7 +100,7 @@ class App extends React.Component<{
         schemes = installedTheme.schemes;
         activeScheme = installedTheme.activeScheme;
       } else {
-        console.log("No theme set as installed");
+        // console.log("No theme set as installed");
```
theRealPadster, line 103: Can we change most of these logs to console.debug instead of commenting them out? They are useful for debugging, and changing to debug would hide them in the console by default.

theRealPadster (changes requested):

src/extensions/extension.tsx
Tetrax-10: its your wish. I just commented it so dev tools wont look clumsy

src/app.tsx
Tetrax-10: Ofcourse feel free to add `console.debug` where ever you want.

src/logic/Utils.ts
rxri: It is removed
image

rxri: I believe he meant whole `else` statement as well

Tetrax-10: > I believe he meant whole `else` statement as well

He just pinged that single line
image

theRealPadster: I meant all the red stuff in the `else` block. I just did a single line because I was on my phone. I didn't realise we weren't removing the default `colors.css` anymore. I think it's fine then, my bad.

Tetrax-10: On your phone? you could have added a line stating `else` block. I would have explained clearly lol

theRealPadster: I thought the "why was this removed" implied the stuff that was removed. ¯\_(ツ)_/¯

Tetrax-10:
LGTM

theRealPadster (approved):

PR #442, opened 2023-02-21
feat: add `category` param to `Marketplace.reset()`

theRealPadster:
This allows you to specify a category of item to reset. It still reloads the window after the operation has completed. I wasn't sure which param format would be best, so I just went with simple one operation per call for now. 

Current usage: 
```js
Marketplace.reset();
Marketplace.reset('extensions');
```

Other potential options we might consider:
```js
Marketplace.reset({
  snippets: true,
});
Marketplace.reset(['theme']);
```

EDIT: We went with a rest param, like so:
```js
// Reset individual items
Marketplace.reset('theme');
Marketplace.reset('theme', 'snippets');
// Reset everything (including settings)
Marketplace.reset();
```

### DIFF
--- a/src/logic/Utils.ts
+++ b/src/logic/Utils.ts
@@ -190,7 +190,7 @@ export const generateSchemesOptions = (schemes: SchemeIni) => {
  * Reset Marketplace localStorage keys
  * @param categories The categories to reset. If none provided, reset everything.
  */
-export const resetMarketplace = (...categories: ("extensions" | "snippets" | "theme" | "all")[]) => {
+export const resetMarketplace = (...categories: ("extensions" | "snippets" | "theme")[]) => {
   console.debug("Resetting Marketplace");
 
   const keysToRemove: string[] = [];
--- a/src/logic/Utils.ts
+++ b/src/logic/Utils.ts
@@ -186,16 +186,41 @@ export const generateSchemesOptions = (schemes: SchemeIni) => {
   ));
 };
 
-// Reset any Marketplace localStorage keys (effectively resetting it completely)
-export const resetMarketplace = () => {
+/**
+ * Reset Marketplace localStorage keys
+ * @param category The category to reset. If not provided, reset all categories.
+ */
+export const resetMarketplace = (category: "extensions" | "snippets" | "theme" | "all" = "all") => {
   console.debug("Resetting Marketplace");
 
-  // Loop through and reset marketplace keys
-  Object.keys(localStorage).forEach((key) => {
-    if (key.startsWith("marketplace:")) {
-      localStorage.removeItem(key);
-      console.debug(`Removed ${key}`);
-    }
+  const keysToRemove: string[] = [];
+
+  // If have category, just reset that category
+  if (category === "extensions") {
+    // Remove the extensions themselves
+    keysToRemove.push(...getLocalStorageDataFromKey(LOCALSTORAGE_KEYS.installedExtensions, []));
+    // Remove the list of extension keys
+    keysToRemove.push(LOCALSTORAGE_KEYS.installedExtensions);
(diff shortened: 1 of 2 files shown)


src/logic/Utils.ts
```diff
@@ -186,16 +186,41 @@ export const generateSchemesOptions = (schemes: SchemeIni) => {
...
+    keysToRemove.push(...getLocalStorageDataFromKey(LOCALSTORAGE_KEYS.installedExtensions, []));
+    // Remove the list of extension keys
+    keysToRemove.push(LOCALSTORAGE_KEYS.installedExtensions);
+  } else if (category === "snippets") {
+    keysToRemove.push(...getLocalStorageDataFromKey(LOCALSTORAGE_KEYS.installedSnippets, []));
+    keysToRemove.push(LOCALSTORAGE_KEYS.installedSnippets);
+  } else if (category === "theme") {
+    keysToRemove.push(...getLocalStorageDataFromKey(LOCALSTORAGE_KEYS.installedThemes, []));
+    keysToRemove.push(LOCALSTORAGE_KEYS.installedThemes);
+    keysToRemove.push(LOCALSTORAGE_KEYS.themeInstalled);
```
kyrie25, line 210: Why do we have 2 of these lol

src/logic/Utils.ts
```diff
@@ -186,16 +186,41 @@ export const generateSchemesOptions = (schemes: SchemeIni) => {
...
-  // Loop through and reset marketplace keys
-  Object.keys(localStorage).forEach((key) => {
-    if (key.startsWith("marketplace:")) {
-      localStorage.removeItem(key);
-      console.debug(`Removed ${key}`);
-    }
+  const keysToRemove: string[] = [];
+
+  // If have category, just reset that category
+  if (category === "extensions") {
```
kyrie25, line 199: Refactoring this to `switch ... case` should make it look visually better imo

src/logic/Utils.ts
```diff
@@ -186,16 +186,41 @@ export const generateSchemesOptions = (schemes: SchemeIni) => {
   ));
 };
 
-// Reset any Marketplace localStorage keys (effectively resetting it completely)
-export const resetMarketplace = () => {
+/**
+ * Reset Marketplace localStorage keys
+ * @param category The category to reset. If not provided, reset all categories.
+ */
+export const resetMarketplace = (category: "extensions" | "snippets" | "theme" | "all" = "all") => {
```
kyrie25, line 193: I think it should be possible to make it accept multiple categories using spread operator.
Just in case like they want to reset 2/3 categories and shouldn't need to type twice.

src/logic/Utils.ts
theRealPadster: Because it has the standard array like the others use, and then it's got a second key that just has the one, since the app references the active theme in a number of places. We can refactor that later if we want; I don't feel like changing how it works in this PR.

src/logic/Utils.ts
theRealPadster: Switch statements always look gross imo, with the extra break and weird indentation and colon syntax. They're also just longer.

src/logic/Utils.ts
kyrie25: That makes sense yeah.

src/logic/Utils.ts
kyrie25: Agree to disagree tbh. That Python-ish approach looks nice to me.

It's generally better practice also since we're doing multiple case comparisons of the same data type.

src/logic/Utils.ts
theRealPadster: Updated.

kyrie25 (approved):

PR #454, opened 2023-03-12
fix: use CLI placeholder value

kyrie25:
Some of these values are a bit off from the default ones (`button` and `button-active`), plus if for some reason these values changes we would only need to commit it from the CLI without needing to do it here also

### DIFF
--- a/resources/color.ini
+++ b/resources/color.ini
@@ -1,1 +1,2 @@
 [Marketplace]
+
--- a/resources/color.ini
+++ b/resources/color.ini
@@ -1,16 +1,1 @@
 [Marketplace]
-text               = ffffff
-subtext            = b3b3b3
-main               = 121212
-sidebar            = 000000
-player             = 212121
-card               = 282828
-shadow             = 000000
-selected-row       = ffffff
-button             = 1ed760
-button-active      = 1ed760
-button-disabled    = 535353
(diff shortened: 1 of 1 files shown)


theRealPadster (approved):
Does the cli automatically inject the right ones and we don't need to define anything to use default? It does seem to work (i.e. look normal) locally if I remove all the definitions and uninstall anything from marketplace.

resources/color.ini
```diff
@@ -1,16 +1 @@
...
-card               = 282828
-shadow             = 000000
-selected-row       = ffffff
-button             = 1ed760
-button-active      = 1ed760
-button-disabled    = 535353
-tab-active         = 333333
-notification       = 2e77d0
-notification-error = cd1a2b
-misc               = 7f7f7f
```
theRealPadster, line 16: Oh, just add the missing newline at eof

theRealPadster (changes requested):

kyrie25:
> Does the cli automatically inject the right ones and we don't need to define anything to use default? It does seem to work (i.e. look normal) locally if I remove all the definitions and uninstall anything from marketplace. 

Yeah that's the idea, the CLI automatically fills in the missing values, so we only need to manage it in one place.

theRealPadster (approved):

PR #456, opened 2023-03-13
chore: remove API request on search `Enter`

kyrie25:
More issues than it is actually useful

### DIFF
--- a/src/components/Grid.tsx
+++ b/src/components/Grid.tsx
@@ -97,11 +97,11 @@ class Grid extends React.Component<
     return installedTheme;
   }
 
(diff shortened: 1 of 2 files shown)


theRealPadster (approved):

theRealPadster:
Only note I would have would be what if all the content hasn't loaded yet and it's just filtering on what's available? Then it might miss some stuff

kyrie25:
> Only note I would have would be what if all the content hasn't loaded yet and it's just filtering on what's available? Then it might miss some stuff

State is updated every time a card is fetched, thus it would still go through the filter again and load if it meets the search query.

PR #459, opened 2023-03-17
fix(readme): set img relative to readme

harbassan:
Fixes the image urls being wrong for all of the themes from the spictify-themes repo. spicetify/spicetify-themes#901


kyrie25:
> When the link starts with a /, it is relative to the root of the repository (regardless of whether the markdown file is nested in subdirectories)

Does this cover for this case as well?

kyrie25 (approved):

PR #467, opened 2023-03-26
feat: add "version" section to bottom of settings modal

theRealPadster:
Adds "version" section to the bottom of settings modal. 
Also adds a few new language strings. 

I'm not sure if this UI is inherently better than just putting "v0.8.5" in the bottom corner somewhere and copying on click or something. Feel free to chime in. 

It would also be nice if we could integrate the update checker here and show a button if there's a new version. 

image

### DIFF
--- a/src/components/Modals/Settings/index.tsx
+++ b/src/components/Modals/Settings/index.tsx
@@ -22,10 +22,18 @@ const SettingsModal = ({ CONFIG, updateAppConfig } : Props) => {
   // then when updating anything in the main state, also updates the SettingsModal state
 
   const [modalConfig, setModalConfig] = React.useState({ ...CONFIG });
+  const [versionButtonText, setVersionButtonText] = React.useState(t("settings.versionBtn"));
   // TODO: use React.useCallback?
   const updateConfig = (CONFIG: Config) => {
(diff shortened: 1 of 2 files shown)


src/components/Modals/Settings/index.tsx
```diff
@@ -75,10 +75,25 @@ const SettingsModal = ({ CONFIG, updateAppConfig } : Props) => {
...
+        </div>
+      </div>
     </div>
   );
 };
 
+const copyVersion = () => {
+  Spicetify.Platform.ClipboardAPI.copy(MARKETPLACE_VERSION);
+  // TODO: It looks like the notification is underneath the modal, so is not visible
+  Spicetify.showNotification(t("settings.versionCopied"));
```
kyrie25, line 94: We could change the text on the Copy button instead yeah?

theRealPadster: Good call. Changed to do this. It'll change back after 3s, which I think makes sense?

kyrie25: That or no timeout at all, since I think it's pretty clear for the user that they copied the text already.

kyrie25 (approved):

Issue #465, opened 2023-03-21
Redesign Settings dialog


theRealPadster:
### 📝 Provide a description of the new feature

- It's super tall and scrolls, which may cause people to not notice there is more than what's initially visible
- The tabs being controlled by toggles and up/down arrows is a weird not particularly intuitive design pattern
- It just doesn't look great
- We could also add the Marketplace version somewhere in here

### ➕ Additional Information

Current implementation (too tall and scrolls on most screens)
![image](https://user-images.githubusercontent.com/1565883/226744916-c9af81b6-3183-4916-82f2-9c2ac750e2f8.png)

Proposed edit to tabs:
What if we make the tabs drag and drop, and click to toggle? We can grey them out or add an outline for enabled ones or something.
![image](https://user-images.githubusercontent.com/1565883/226744808-8816af4e-b552-41a0-93ac-61b156075445.png)

We could also use an accordion for the sections. e.g. "Options" starts collapsed, but you can click it and it expands to show the settings you can toggle.

theRealPadster:
Alternatively, we could do something like Finder's toolbar customization (I believe Firefox has similar functionality for its toolbar). We would two sections, one with inactive tabs and one with active tabs, and you can drag and drop them around to order or remove. 

<img width="711" alt="image" src="https://user-images.githubusercontent.com/1565883/227809306-be468895-de45-4246-a634-7de9f969e1ad.png">

theRealPadster:
I don't think this was meant to be closed with the commit that cleaned it up a bit. It still could be much better and not all of the things here have been addressed.

zhngharry:
I can work on this, is this still open?

theRealPadster:
> I can work on this, is this still open?

I think it could still look a bit better, so sure

PR #468, opened 2023-03-26
feat: settings modal styling

theRealPadster:
This PR can be for redesigning the settings modal. Adding dividers like Nord Spotify has seems to help clean things up a little bit, but it would be nice to simplify things more as outlined in #465 

This is what I have so far. 
image


kyrie25 (approved):
Looks p good

theRealPadster:
> Looks p good

I'm thinking of using [this library](https://www.npmjs.com/package/react-drag-reorder) to add drag to reorder support for the tabs. Do you think that makes sense to put in this PR or a separate one?

kyrie25:
> > Looks p good
> 
> 
> 
> I'm thinking of using [this library](https://www.npmjs.com/package/react-drag-reorder) to add drag to reorder support for the tabs. Do you think that makes sense to put in this PR or a separate one? 

Since we reworking the modal in the first place I think it's better to do it on this one.

theRealPadster:
I'm just gonna revert the drag and drop thing for now so I can get this PR done. We can try it again later if we want.

theRealPadster:
Looks like this now. Is that fine? (In two parts because it's tall)

image
image

rxri (approved):

Issue #435, opened 2023-02-07
Add user created snippets


theRealPadster:
### 📝 Provide a description of the new feature

It would be sweet to add user created snippets to marketplace, so it's easier to get them added in, and we don't need people to make a PR and stuff every time they want to add/update one. 

This would work much the same as the themes and extensions tabs, but have a `snippetcss` key or something and use a snippet github topic.

theRealPadster:
I don't remember. I'm proposing we just use a manifest like all the rest

theRealPadster:
Should we make the manifest have a `snippetcss` field that _is_ the css string, or have it point to a css file, like the usercss one does? Pointing to a file would let people keep things cleaner if they have more than one snippet, and would allow people to use less or scss and build the css itself. But also might just be overkill for snippets.

CharlieS1103:
@theRealPadster I forget why we didn't allow this from the start, did we have a good reason?

CharlieS1103:
The only good reason I can come up with now is to prevent the saturation of unmaintained snippets, but at this point I think most of them are like that regardless.

theRealPadster:
@CharlieS1103 tbh I think we just didn't think of it? Like we just added them ourselves and didn't really think of other people adding their own in their own repo.

CharlieS1103:
@theRealPadster thats very valid on past us's behalf. I'll look into merging that PR

PR #478, opened 2023-04-10
feat(snippets): add more rounded corners

Fadexz:
Adds some changes to round some extra elements.


theRealPadster (approved):
@Fadexz It looks like there were merge conflicts. I think your branch may not be up to date with upstream, since the "Rounded Images" snippet was added. Can you just double check that I merged it properly and the snippet looks as intended? We can merge once you approve.

Fadexz (commented):
Looks good

Fadexz:
Seems for some reason it wasn't in there from before, so there was a conflict it seems.
Does this only take effect when the marketplace has an update and isn't fetched automatically?

theRealPadster:
> Seems for some reason it wasn't in there from before, so there was a conflict it seems. Does this only take effect when the marketplace has an update and isn't fetched automatically?

Yeah I think it was because you had an existing fork that was behind upstream. You need to sync your fork so it's up to date or it will be behind and could have merge conflicts.

Fadexz:
> > Seems for some reason it wasn't in there from before, so there was a conflict it seems. Does this only take effect when the marketplace has an update and isn't fetched automatically?
> 
> Yeah I think it was because you had an existing fork that was behind upstream. You need to sync your fork so it's up to date or it will be behind and could have merge conflicts.

Yes I tried to do that at the time but it seems it didn't work so I just deleted the local copy to make sure it was all good.

If you don't mind, do you know the answer to my question? Do snippets only get added to the marketplace once a new release of the marketplace occurs? I assume that it is since I don't see it, so it would take maybe a month or so?

theRealPadster:
> If you don't mind, do you know the answer to my question? Do snippets only get added to the marketplace once a new release of the marketplace occurs? I assume that it is since I don't see it, so it would take maybe a month or so?
> 
> I think this answers my question though: #435

If I recall correctly, it fetches and caches the snippets from the GitHub repo. So doesn't need a new release, just needs to be merged in. That issue is different. It's proposing we let users put snippets in a manifest file and host them in their own repo, like how the rest of the marketplace works.

Fadexz:
> If I recall correctly, it fetches and caches the snippets from the GitHub repo. So doesn't need a new release, just needs to be merged in. That issue is different. It's proposing we let users put snippets in a manifest file and host them in their own repo, like how the rest of the marketplace works.

I see. I have tried to do that before and have just now to make sure and I even see the lists populate but I don't see the snippet I added. So you can see it just to clarify? If so i'll have to try a bit harder to make sure I clear everything.

theRealPadster:
I have it in my snippets, but it looks like it's the version that hasn't been updated. I took another look at the code and it looks like it's actually getting them from the local file (your local copy of the marketplace custom app) and not pulling from github. So you'd have to download the latest build of marketplace to update it. We should probably change it to do it remotely, if you want to make a new issue for that.

PR #493, opened 2023-05-21
feat(snippets): Rotating Album Cover Fix

illtscha:
After a Spotify update I noticed that the old "Rotating Cover Art" does not work anymore (at least for me). So I wrote a new one that works.


kyrie25 (changes requested):
Please fix the previous snippet instead of making a new one with the exact functionality

illtscha:
Can I add my cover picture for the addon?

kyrie25:
> Can I add my cover picture for the addon?

If the previous one works then there would be no need for it.

src/resources/snippets.ts
```diff
@@ -164,8 +164,8 @@ export default [
   {
     "title": "Rotating Cover Art",
     "description": "Adds circular mask to cover art and rotation",
-    "code": `@keyframes rotating { from { transform: rotate(0deg); } to { transform: rotate(360deg); }} .cover-art { animation: rotating 2s linear infinite; clip-path: circle(50% at 50% 50%);}`,
-    "preview": "resources/assets/snippets/rotating-coverart.png",
+    "code": `@keyframes rotating { from { transform: rotate(0deg); } to { transform: rotate(360deg); }} .FqmFsMhuF4D0s35Z62Js, .zmOtW0vqqn1qpZrtQ_w9 { animation: rotating 10s linear infinite; clip-path: circle(50% at 50% 50%);} .H0HbpIM3UrcupWIAjLWu { background: none; } .H0HbpIM3UrcupWIAjLWu { background: none; } .main-coverSlotCollapsed-container { padding-right: 10px !important; }`,
```
kyrie25, line 167: These classes are already mapped in latest `css-map` commits and the snippet still works as expected.

https://github.com/spicetify/spicetify-marketplace/assets/77577746/57e8c427-fe89-4ec4-9615-93ad2b19454d

Run `spicetify restore backup apply` to fetch the new mappings

kyrie25 (changes requested):

src/resources/snippets.ts
illtscha: oh ok thank you

PR #515, opened 2023-06-08
fix(Grid): render card on tab change

kyrie25:
Cache tab on page load function so that it can be compared later on when rendering cards.
Fix duplicate card rendering on Installed tab and wrong cards being rendered after tab change.


kyrie25:
Also prevent duplicate card rendering when you spam the suitable search query repeatedly by reusing the same key instead of creating a new one
image

src/components/Grid.tsx
```diff
@@ -5,7 +5,7 @@ import { Option } from "react-dropdown";
 const Spicetify = window.Spicetify;
 
 import { CardItem, CardType, Config, SchemeIni, Snippet, TabItemConfig } from "../types/marketplace-types";
-import { getLocalStorageDataFromKey, generateSchemesOptions, injectColourScheme, generateKey } from "../logic/Utils";
```
theRealPadster, line 8: What's it doing here that's different from what the `generateKey` function does? Could we just update `generateKey`?

kyrie25: `generateKey` is depended on by localStorage management. I dare not change that since that would mean breaking theme installation for everyone and removing existing config.

theRealPadster: Ah fair

theRealPadster (approved):

PR #519, opened 2023-06-13
fix(snippets): hover panel - width and back fix

Om-Thorat:


Om-Thorat:
It is completed and fixed

kyrie25 (approved):

PR #545, opened 2023-07-08
feat(grid): sorting options

CharlieS1103:


src/components/Grid.tsx
```diff
@@ -188,27 +188,26 @@ class Grid extends React.Component<
...
           repo.default_branch,
           repo.stargazers_count,
-          this.CONFIG.visual.hideInstalled,
+          true,
         );
 
-        // I believe this stops the requests when switching tabs?
-        if (this.requestQueue.length > 1 && queue !== this.requestQueue[0]) {
-          // Stop this queue from continuing to fetch and append to cards list
-          return -1;
```
theRealPadster, line 202: Do we know for sure what this part did, and that it's okay to take out?

src/components/Grid.tsx
```diff
@@ -250,38 +249,37 @@ class Grid extends React.Component<
...
+        const repoThemes = await fetchThemeManifest(
           repo.contents_url,
           repo.default_branch,
           repo.stargazers_count,
         );
-        // I believe this stops the requests when switching tabs?
-        if (this.requestQueue.length > 1 && queue !== this.requestQueue[0]) {
-          // Stop this queue from continuing to fetch and append to cards list
-          return -1;
-        }
```
theRealPadster, line 264: Same note as above

src/components/Grid.tsx
```diff
@@ -250,38 +249,37 @@ class Grid extends React.Component<
...
 
-        if (themes && themes.length) {
-          themes.forEach((theme) => {
-            Object.assign(theme, { lastUpdated: repo.pushed_at });
-            this.appendCard(theme, "theme", activeTab);
-          });
+        if (repoThemes && repoThemes.length) {
+          themes.push(...repoThemes.map((theme) => ({
+            ...theme, lastUpdated: repo.pushed_at,
+          })));
```
theRealPadster, line 263: I think it was more readable before tbh. Too much spread lol.

src/components/Grid.tsx
```diff
@@ -250,38 +249,37 @@ class Grid extends React.Component<
...
 
-      // First request is null, so coerces to 1
+      sortThemes(themes, localStorage.getItem("marketplace:sort") || "stars");
+
+      for (const theme of themes) {
+        this.appendCard(theme, "theme", activeTab);
+      }
+
       const currentPage = this.requestPage > -1 && this.requestPage ? this.requestPage : 1;
-      // -1 because the page number is 1-indexed
```
theRealPadster, line 276: Doesn't hurt to leave this in

src/logic/Utils.ts
```diff
@@ -579,3 +588,44 @@ export const addExtensionToSpicetifyConfig = (main?: string) => {
...
+    extensions.sort((a, b) => new Date(a.lastUpdated).getTime() - new Date(b.lastUpdated).getTime());
+    break;
+  case "stars":
+  default:
+    extensions.sort((a, b) => b.stars - a.stars);
+    break;
+  }
+};
+// TODO: Create theme type
+export const sortThemes = (themes: CardItem[], sortOption: string) => {
```
theRealPadster, line 612: Aren't these two sort functions the exact same? Why not just use one?

src/components/Grid.tsx
```diff
@@ -250,38 +249,37 @@ class Grid extends React.Component<
...
       const remainingResults = pageOfRepos.total_count - soFarResults;
 
       console.debug(`Parsed ${soFarResults}/${pageOfRepos.total_count} themes`);
       if (remainingResults > 0) return currentPage + 1;
       else console.debug("No more theme results");
       break;
-    } case "Apps": {
+    }
+    case "Apps": {
       const pageOfRepos = await getTaggedRepos("spicetify-apps", this.requestPage, this.BLACKLIST);
```
theRealPadster, line 283: Probably add sort for apps as well, to be consistent. Maybe installed too.

theRealPadster (changes requested):

src/components/Grid.tsx
CharlieS1103: Can't do it with the old way, this is not just a readability change but a functionality change as well.

src/components/Grid.tsx
CharlieS1103: Honestly, no clue what it does, I think you added it a while ago to prevent the issue where cards would appear if you switch tabs while it's still loading, however, after experimenting a lot, I don't think this is an issue anymore for whatever reason? (Perhaps changes with how we fetch stuff?)

src/components/Grid.tsx
CharlieS1103: I think installed would be a bit much but i'll try to do it with apps

src/components/Grid.tsx
theRealPadster, line 263: I think this is slightly easier to read?
```suggestion
        if (repoThemes && repoThemes.length) {
          themes.push(...repoThemes.map(
            (theme) => ({
              ...theme, 
              lastUpdated: repo.pushed_at,
            })
          ));
```

theRealPadster: Fair. Added a separate suggestion for readability.

CharlieS1103:
@theRealPadster Thoughts on merging as is, in terms of the apps and installed tabs to be entirely honest we have so few of them i doubt it'd be fully necessary to implement as of now, plus the systems they use are entirely different to the themes and extensions tabs so it would probably take loads more work to implement

src/components/Grid.tsx
theRealPadster: I've added for both. And snippets, though the star/oldest/newest sorts don't do anything

theRealPadster:
I've done some more work on it. Added support for sort to the rest of the tabs (though stars/oldest/newest don't do anything for snippets). I considered disabling/removing those for the Snippets tab, but that made the sort reset when you switched tabs, which was kind of annoying, so I removed it. I'd be okay if we just remove the sort dropdown on the Snippets tab full stop, but up to you guys. 

I also fixed the header not being sticky when you scroll, and replaced the redundant "spicetify marketplace" text on the left with the sort dropdown and a label. 

image

src/components/Grid.tsx
theRealPadster: I put it back in, just in case 🤷

theRealPadster (approved):
LGTM

PR #550, opened 2023-07-13
fix(install): get userdata path

SunsetTechuila:
make sure that output contains only path


SunsetTechuila:
alternatively could be fixed by https://github.com/spicetify/spicetify-cli/pull/2470

kyrie25 (approved):

PR #569, opened 2023-08-02
feat(snippets): add hide all play counts and popular tracks

greenimpala:


theRealPadster:
Would it make more sense to split this up into two snippets?

greenimpala:
> Would it make more sense to split this up into two snippets?

Popular tracks also contain play counts which I want to hide. There is [already](https://github.com/spicetify/spicetify-marketplace/blob/f00ac236e0a14084d44a3b34f0c63e2501d5bc3e/src/resources/snippets.ts#L267) a snippet to hide play counts but it wasn't appearing in the marketplace tab `2.22.1` 🤔.

theRealPadster:
The hide play counts snippet shows up for me on 2.22.1. 
image
Where can I find the "popular tracks" bit of the UI that this snippet hides? Is it that tightly coupled to the play counts? If not, I still think it would make sense to just install the two snippets to hide both.

rxri (commented):
No activity for 7 months and `Hide play count` snippet already exists in marketplace

PR #572, opened 2023-08-05
feat: use Workers for manifest fetching & optimizations

kyrie25:
- Use Web Workers to fetch manifest contents, this way the console will no longer be littered with 404 spams.
image

- Reworked & fixed cache system (so that it actually works now), no longer making duplicate requests and decrease load time.

- Topic results will also be cached and will no longer cause rate limits when spamming switching tabs.

- A notification will now appear if all JSDelivr endpoints can't be connected to (except when they are offline, in which the extension will reload when online)
image


src/logic/FetchRemotes.ts
```diff
@@ -83,7 +98,7 @@ export async function fetchExtensionManifest(contents_url: string, branch: strin
     if (!regex_result || !regex_result.groups) return null;
     const { user, repo } = regex_result.groups;
 
-    manifests = await getRepoManifest(user, repo, branch);
+    manifests = await getRepoManifest(user, repo, regex_result[0]);
```
theRealPadster, line 101: Why is this using regex result instead of branch?

src/logic/FetchRemotes.ts
```diff
@@ -130,8 +145,8 @@ export async function fetchExtensionManifest(contents_url: string, branch: strin
 
     return parsedManifests;
   }
-  catch (err) {
-    // console.warn(contents_url, err);
+  catch (err: any) {
```
theRealPadster, line 148: Types

src/types/marketplace-types.d.ts
```diff
@@ -162,3 +162,25 @@ export type Config = {
...
+    self: string;
+    git: string;
+    html: string;
+  };
+};
+
+export type GithubMessage = {
+  message: string;
+  documentation_url: string;
+};
```
theRealPadster, line 186: Missing EOF newline

theRealPadster (changes requested):

src/logic/FetchRemotes.ts
kyrie25: You can only assign either unknown or any to caught errors and type check is done in the catch block anyway.

src/logic/FetchRemotes.ts
```diff
@@ -51,17 +83,20 @@ export async function getTaggedRepos(tag: RepoTopic, page = 1, BLACKLIST:string[
 * @returns The manifest object
 */
 async function getRepoManifest(user: string, repo: string, branch: string) {
-  const sessionStorageItem = window.sessionStorage.getItem(`${user}-${repo}`);
-  const failedSessionStorageItems = window.sessionStorage.getItem("noManifests");
+  const key = `${user}-${repo}`;
```
theRealPadster, line 86: Do we not need the branch anymore? What if the user has one repo and uses different branches for different projects?

kyrie25: That's still the same key we used before, just initialized once.
Also one repo only has one manifest. This caches everything in the manifest file, not one manifest object itself.

src/components/Card/Card.tsx
```diff
@@ -379,7 +378,7 @@ class Card extends React.Component<CardProps, {
    */
   async fetchAndInjectUserCSS(theme) {
     try {
-      const tld = await getAvailableTLD();
+      const tld = window.sessionStorage.getItem("tld") || "";
```
rxri, line 381:
```suggestion
      const tld = window.sessionStorage.getItem("marketplace-request-tld") || undefined;
```

src/logic/Utils.ts
```diff
@@ -417,9 +417,11 @@ export const initAlbumArtBasedColor = (scheme: ColourScheme) => {
   });
 };
 
-export const parseCSS = async (themeData: CardItem, tld = "net") => {
+export const parseCSS = async (themeData: CardItem, tld?: string) => {
```
rxri, line 420:
```suggestion
export const parseCSS = async (themeData: CardItem, tld?: string | undefined) => {
```

rxri (changes requested):

src/logic/Utils.ts
kyrie25: Optional param is explicitly undefined

theRealPadster (approved):

rxri (approved):

PR #577, opened 2023-08-19
feat(snippets): add fix listening on and default progress bar snippets

Toomas633:


Toomas633:
For themes that move the progress bar to the bottom and for some themes move the listening on outside the now playing bar. Compatible with theme coloring. Tested on Comfy theme.

theRealPadster:
It looks like this adds a new locale as well? Would you be able to put that as a separate PR please?

Toomas633:
My bad forgot to change branches

theRealPadster (approved):

PR #578, opened 2023-08-19
feat(snippets): smaller now playing bar

ajtruex:
makes the now playing bar at the bottom smaller
smaller-now-playing-bar
Code
`.main-nowPlayingBar-nowPlayingBar {height: auto}`


theRealPadster (changes requested):
Hey sorry, just got around to testing this. It looks like it makes the progress bar slightly shorter vertically? Might there be a more descriptive name for the snippet that clarifies it's the height that it's changing, not the width? 

I was also going to resolve the merge conflicts in the web editor, but it said it would commit to main. This may mean your main branch or the origin spicetify main branch, but I wasn't sure, so I cancelled it. Could you pull in the changes and resolve the conflicts on your end? It looks like it's just a couple other snippets that have been added since your code was committed.

rxri:
No activity and none of request changes have been implemented

PR #582, opened 2023-08-26
feat(i18n): add estonian translation

Toomas633:


src/app.tsx
```diff
@@ -29,6 +30,8 @@ i18n
       ru,
       "zh-TW": zhTW,
       "zh-CN": zhCN,
+      et,
+      "et-EE": et,
```
theRealPadster, line 34: Does Spotify give an option for both of these locales? I think if you just add it once, it should use it as a fallback if the corresponding one is in use?

Toomas633: removed

theRealPadster (approved):
Looks good, thanks!

PR #583, opened 2023-08-27
refactor(Grid): move setState outside of loops to avoid unnecessary redraws

Delusoire:
The previous implementation was constantly re-rendering the Grid whenever a card was added, causing terrible performance. This PR aims to solve this by only redrawing once all cards are added.


theRealPadster (approved):
Nice, seems to be much faster now, thanks!

PR #594, opened 2023-09-03
feat(snippets): smooth progress/volume bar

Droidiar:
Makes the Progress/Volume bar smooth!

using a very very simple method :3

preview
 the gif is very very slow


theRealPadster (changes requested):
This seems good to me but what is that icon? Is there anything else that would be more indicative of what the snippet does?

Droidiar:
If I was able to add an animated media I could probably just use the gif I sent but I could make something better if that's out of the picture. 
I was kinda workin with just a 50x50 picture so its bad yeah.

theRealPadster:
@Droidiar How is this?
smooth-progress-bar

Droidiar:
Oh! Thank you, that's really good.

theRealPadster (approved):

PR #605, opened 2023-10-03
fix(snippets): oneko and duck

SunsetTechuila:
before:
image

after:
image


theRealPadster:
This would be better served as a fix to the actual repos for the oneko and duck extensions. Might as well fix the problem at the source instead of adding snippets for things that don't even apply to the base install.

SunsetTechuila:
> oneko and duck extensions

there is no duck extension

> fix the problem at the source instead

oneko doesn't have this issue

> adding snippets

I didn't add anything, I fixed already added snippets

> things that don't even apply to the base install

they do, these snippets are completely standalone

theRealPadster (approved):

PR #610, opened 2023-10-11
fix(grid styles): header position and background

SunsetTechuila:
before:

image

image

after:

image

image

### DIFF
--- a/src/styles/components/_grid.scss
+++ b/src/styles/components/_grid.scss
@@ -13,7 +13,7 @@
 
   // To position the settings button + colour schemes
   position: sticky;
-  flex-direction: row-reverse;
+  flex-direction: row;
   top: 64px;
   z-index: 1;
   
@@ -29,11 +29,6 @@
   & > :is(div, button) {
     margin-left: 8px;
   }
-}
-
-.marketplace-header__left {
-  position: absolute;
-  left: 0;
 }
 
 .marketplace-header__label {
--- a/src/styles/components/_grid.scss
+++ b/src/styles/components/_grid.scss
@@ -9,16 +9,18 @@
   background-color: var(--spice-main);
   display: flex;
   justify-content: space-between;
-  height: 64px;
+  padding-bottom: 8px;
 
   // To position the settings button + colour schemes
   position: sticky;
   flex-direction: row-reverse;
   top: 64px;
   z-index: 1;
-}
-.contentSpacing:has(#marketplace-readme) .marketplace-header {
-  top: 0;
+  
+  // Fix positioning on readme pages
+  .contentSpacing:has(#marketplace-readme) & {
+    top: 0;
+  }
 }
 
 .marketplace-header__right,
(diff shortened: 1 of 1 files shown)


theRealPadster (approved):
Ah nice, I think this is what I was going for originally. Thanks for catching the readme bug too 🙃

theRealPadster:
> I don't think sticky header is a good idea tho

I personally found it really annoying to have to scroll all the way up to the top to change colour schemes or search. Maybe we can tighten the spacing up a bit so it's not as tall?

src/styles/components/_grid.scss
```diff
@@ -6,16 +6,20 @@
   align-content: space-between;
   align-items: center;
   color: var(--spice-text);
+  background-color: var(--spice-main);
   display: flex;
   justify-content: space-between;
-  margin: 16px 0;
+  height: 64px;
```
theRealPadster, line 12:
```suggestion
  padding-bottom: 8px;
```

src/styles/components/_grid.scss
```diff
@@ -6,16 +6,20 @@
...
 
   // To position the settings button + colour schemes
   position: sticky;
   flex-direction: row-reverse;
-  top: 80px;
+  top: 64px;
   z-index: 1;
```
theRealPadster, line 18:
```suggestion
  // Fix positioning on readme pages
  .contentSpacing:has(#marketplace-readme) & {
    top: 0;
  }
  z-index: 1;
```

src/styles/components/_grid.scss
```diff
@@ -6,16 +6,20 @@
...
 }
+.contentSpacing:has(#marketplace-readme) .marketplace-header {
+  top: 0;
+}
```
theRealPadster, line 22:
```suggestion
  top: 64px;
  z-index: 1;
  
  // Fix positioning on readme pages
  .contentSpacing:has(#marketplace-readme) & {
    top: 0;
  }
}
```

theRealPadster:
I would also remove this part. It doesn't let me suggest on unmodified lines I guess. 
```css
.marketplace-header__left {
  position: absolute;
  left: 0;
}
```

theRealPadster:
Screenshot for reference
image

SunsetTechuila:
> I would also remove this part.

Then the right and left elements will swap places

image

SunsetTechuila:
> I would also remove this part. It doesn't let me suggest on unmodified lines I guess.

done

theRealPadster (approved):

AliahX:
This seems to have caused another issue?
The marketplace header is now displayed on top of the page content even when scrolled all the way to the top of the page, obscuring headers and anything else at the top of the page.
Here is a screenshot showing the issue
Screenshot 2024-06-19 173048
image
This is a fix I made by messing around with the css a bit, but it affected non spicetify pages too.
Screenshot 2024-06-19 173210

AliahX:
You're right yeah. idk why I sent this in the marketplace repo...

PR #611, opened 2023-10-12
feat(grid): sort by creation date

SunsetTechuila:
the "newest" and "oldest" sorting options are actually responsible for sorting by last update date, which doesn't match their names, so I renamed them to "recently updated" and "most stale" and added sorting by repository creation date

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -40,6 +40,7 @@ class Card extends React.Component<CardProps, {
   tagsExpanded: boolean;
   externalUrl: string;
   lastUpdated: string | undefined;
+  created: string | undefined;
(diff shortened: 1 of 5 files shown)


theRealPadster (changes requested):
Good idea. Could we change it to be "last updated" though, instead of "recently updated"?

SunsetTechuila:
how do I name the variable then? there is already "lastUpdated" variable
https://github.com/spicetify/spicetify-marketplace/pull/611/files#diff-c8e6101bb10d5263d0ca774fd2a5696106c7740789e43673f74f845bbcc67152R94

theRealPadster (approved):

PR #612, opened 2023-10-12
feat(grid): remove unsupported sorting options on snippets page

SunsetTechuila:


src/components/Grid.tsx
```diff
@@ -546,11 +546,16 @@ class Grid extends React.Component<
...
-              sortBySelectedFn={(a) => a.key === this.CONFIG.sort} />
+            {this.CONFIG.activeTab !== "Snippets" ? (
+              <>
+                <h2 className="marketplace-header__label">{t("grid.sort.label")}</h2>
+                <SortBox
+                  onChange={(value) => this.updateSort(value)}
+                  sortBoxOptions={generateSortOptions(t)}
+                  sortBySelectedFn={(a) => a.key === this.CONFIG.sort} />
+              </>
+            ) : <h1>{this.props.title}</h1>
```
theRealPadster, line 557: Is the `this.props.title` the "Spicetify Marketplace text? I removed that in the original PR.

theRealPadster:
The sort does work on the snippets page, it just only supports sorting by name. I'm not sure what the best solution is here, since I can see it being useful as there are a lot of snippets now.

src/components/Grid.tsx
SunsetTechuila: > I removed that in the original PR

I know, but without it header looks empty

theRealPadster: Still, doesn't make sense to have a title only show up on a single tab.

SunsetTechuila:
> it just only supports sorting by name

does it? then please wait

SunsetTechuila:
oh, now it resets

SunsetTechuila:
ok, then I don't know what to do without creating kludges

PR #613, opened 2023-10-13
feat(install/windows): use provided userdata path

SunsetTechuila:
brings support for scoop/winget/etc spicetify installations, but will still work if spicetify returns error/warning with/without path - no cons

### DIFF
--- a/resources/install.ps1
+++ b/resources/install.ps1
@@ -12,7 +12,11 @@ if (-not (Get-Command -Name spicetify -ErrorAction SilentlyContinue)) {
   Invoke-WebRequest @Parameters | Invoke-Expression
 }
 
-$spiceUserDataPath = "$env:APPDATA\spicetify"
+spicetify path userdata | Out-Null
+$spiceUserDataPath = (spicetify path userdata)
+if (-not (Test-Path -Path $spiceUserDataPath -PathType Container)) {
+  $spiceUserDataPath = "$env:APPDATA\spicetify"
+}
 $marketAppPath = "$spiceUserDataPath\CustomApps\marketplace"
 $marketThemePath = "$spiceUserDataPath\Themes\marketplace"
 $isThemeInstalled = $(


theRealPadster:
I installed WingetUI on my PC, to make it slightly more like a proper package management system, but when I try and update Spicetify via there, it fails and gives me this error:
```
Successfully verified installer hash
Extracting archive...
Successfully extracted archive
Starting package install...
Unable to remove Portable package as it has been modified; to override this check use --force
```

This is supposed to fix that somehow? Is there some way I can test it? I don't use powershell at all and have no idea how this stuff works lol.

SunsetTechuila:
> This is supposed to fix that somehow?

No, scoop (definitely) and winget (probably) install spicetify to different locations and spicetify uses different folders to store user data, so when the script installs marketplace to appdata/roaming it fails to apply. So the main point is to fix this by calling spicetify to get the user data path instead of using the hardcoded path

SunsetTechuila:
may be relevant for chocolatey as well

theRealPadster:
Ah, this is the marketplace install script, I was thinking it was the main one. I can probably test it later today then. Sounds like a good idea to me

resources/install.ps1
```diff
@@ -12,7 +12,11 @@ if (-not (Get-Command -Name spicetify -ErrorAction SilentlyContinue)) {
   Invoke-WebRequest @Parameters | Invoke-Expression
 }
 
-$spiceUserDataPath = "$env:APPDATA\spicetify"
+spicetify path userdata | Out-Null
```
theRealPadster, line 15: I don't really understand PowerShell at all, but is this line doing anything? Doesn't look like it's assigning any variable or anything

SunsetTechuila: #550

theRealPadster: It's just outputting it to console?

SunsetTechuila: calls without output

SunsetTechuila: or you are speaking about "config has been generated" message?

theRealPadster (approved):

PR #617, opened 2023-10-16
feat: migrate to pnpm

SunsetTechuila:


.github/workflows/push.yml
```diff
@@ -3,42 +3,50 @@ on:
...
     name: Push
     runs-on: ubuntu-latest
+    env:
+      TARGET_BRANCH: dist
+    permissions:
+      contents: write
     steps:
-      - name: Check out Git repository
-        uses: actions/checkout@v2.4.0
+      - name: Checkout Project
```
theRealPadster, line 16:
```suggestion
      - name: Check Out Project
```

theRealPadster:
Is this all tested and good to go now?

.github/workflows/push.yml
rxri: checkout is correct though

SunsetTechuila:
don't we need to add all yarn and npm files to gitignore or only allow pnpm?

theRealPadster:
> don't we need to add all yarn and npm files to gitignore or only allow pnpm?

I don't think we need to, but it would be a good idea, yes

SunsetTechuila:
What exactly is a good idea? Add to gitignore or allow only pnpm?

theRealPadster:
Add the files to `.gitignore`, so people don't accidentally commit them. I'm not sure how you'd only allow pnpm.

SunsetTechuila:
packageManager property doesn't force to use it

rxri:
and use `engines` and `engine-strict` option in .npmrc

PR #620, opened 2023-10-21
feat: home header color change

aeginyma:


resources/assets/snippets/preview-home-header.png
aeginyma, line 1: Old snippet preview!

rxri:
No activity, there was failing linting since the beginning. Snippet did not allow to set custom hex color and required user to copy it and change it

PR #638, opened 2023-11-03
chore(package.json): add `packageManager`

SunsetTechuila:
> version: latest

not needed if the version is specified in package.json
[link](https://github.com/pnpm/action-setup/tree/master?tab=readme-ov-file#version)

> "node": ">=18"

there is no reason to have 20 here, check the dependencies. npm and pnpm only show a warning, but yarn throws an error if you have the node ver less than 20

> "build:prod": "npm run build:local && npm run copy:docs"

why have pnpm here if we don't force to use it?


package.json
```diff
@@ -9,10 +9,11 @@
 	"bugs": {
 		"url": "https://github.com/spicetify/spicetify-marketplace/issues"
 	},
+	"packageManager": "pnpm@8.10.2",
 	"scripts": {
 		"build": "spicetify-creator",
 		"build:local": "spicetify-creator --out=dist --minify",
-		"build:prod": "pnpm build:local && pnpm copy:docs",
+		"build:prod": "npm run build:local && npm run copy:docs",
```
theRealPadster, line 16: I'm confused. Why are we changing this to npm?

package.json
```diff
@@ -21,7 +22,7 @@
 		"prepare": "husky install"
 	},
 	"engines": {
-		"node": ">=20"
+		"node": ">=18"
```
theRealPadster, line 25: And why downgrade to 18?

SunsetTechuila:
@theRealPadster did you read the pr description?

theRealPadster:
> @theRealPadster did you read the pr description?

Yes. It sounded to me like you were going to make the scripts use pnpm, but then I looked and it's changing it back to npm

package.json
SunsetTechuila: > there is no reason to have 20 here, check the dependencies. npm and pnpm only show a warning, but yarn throws an error

package.json
SunsetTechuila: > why have pnpm here if we don't force to use it?

theRealPadster: Are we not forcing to use it? I don't get it

package.json
theRealPadster: You mean use an older version because we don't need anything newer? What's wrong with just using the newest?

package.json
SunsetTechuila: > Are we not forcing to use it?

yes, we don't force to use pnpm

package.json
SunsetTechuila: I'm using 18 and I don't want to upgrade without the reason

SunsetTechuila: You can build the Marketplace even with node 12

theRealPadster: Just use NVM, I probably have like 6 node versions installed

.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
theRealPadster, line 1:
```suggestion
engine-strict=true

```

package.json
```diff
@@ -21,6 +22,9 @@
 		"prepare": "husky install"
 	},
 	"engines": {
+		"yarn": "please-use-pnpm",
+		"npm": "please-use-pnpm",
```
theRealPadster, line 26: lol is this a thing?

theRealPadster (changes requested):

package.json
rxri: it is

theRealPadster (approved):

PR #643, opened 2023-11-05
fix(workflows): auto-release

SunsetTechuila:
i'm really sorry

### DIFF
--- a/.github/workflows/auto_release.yml
+++ b/.github/workflows/auto_release.yml
@@ -25,7 +25,7 @@ jobs:
         id: check_result
         run: |
           latest_release_version=$(git for-each-ref --sort=-creatordate --count 1 --format="%(refname:short)" "refs/tags/*")
-          current_version=$(node -p "require('./package.json').version")
+          current_version="v$(node -p "require('./package.json').version")"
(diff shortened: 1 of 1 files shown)


.github/workflows/auto_release.yml
```diff
@@ -54,6 +54,6 @@ jobs:
       - name: Publish the release
         uses: softprops/action-gh-release@v1
         with:
-          tag_name: "v${{ needs.check_version.outputs.new_version }}"
+          tag_name: ${{ needs.check_version.outputs.new_version }}
```
theRealPadster, line 57: No worries my dude. Don't these need to match though? Looks like it's checking current_version with the v tag and setting the new one to have no v?

theRealPadster: Or do you mean just do this once and then we change it to just not have the v at all?

SunsetTechuila: No, right now both variables have "v" before the version number

theRealPadster: Ohh yes I see, my bad.

theRealPadster (approved):

PR #647, opened 2023-11-05
feat: add update window

### LINKED ISSUE #645: Show an upgrade modal, like Spicetify itself does
### 📝 Provide a description of the new feature

Have a modal pop up when you click the "download new version" button that gets added to the toolbar if there's a new Marketplace version. 

It could have the changelog/release notes, or even just a title, link, and note about how to update marketplace. 

### ➕ Additional Information

For reference:
image

SunsetTechuila:
resolves #645


SunsetTechuila:
image

src/styles/styles.scss
```diff
@@ -6,12 +6,16 @@
...
-@import url('./components/_grid.scss');
-@import url('./components/_dropdown.scss');
-@import url('./components/_card.scss');
-@import url('./components/_settings.scss');
-@import url('./components/_reload-modal.scss');
-@import url('./components/_add-snippet-modal.scss');
-@import url('./components/_readme-pages.scss');
-@import url('./components/_fixes.scss');
-@import url('./components/_import-export.scss');
+@use './components/prismjs-themes/prism-tomorrow.scss';
```
theRealPadster, line 9: Ahh, this works now? I guess we were just doing it wrong?

.eslintrc.js
```diff
@@ -12,11 +12,9 @@ module.exports = {
   ],
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
-    "ecmaFeatures": {
-      "jsx": true,
```
theRealPadster, line 16: We don't need this anymore?

.eslintrc.js
```diff
@@ -1,9 +1,9 @@
-// eslint-disable-next-line no-undef
 module.exports = {
-  "ignorePatterns": ["node_modules", "dist"],
+  "ignorePatterns": ["node_modules", "dist", ".eslintrc.js", "spicetify.d.ts"],
   "env": {
     "browser": true,
-    "es2021": true,
+    "commonjs": true,
```
theRealPadster, line 5: Do we need commonjs? We don't use `require()`

SunsetTechuila: It is not only for "require"

.eslintrc.js
SunsetTechuila: We don't

src/styles/styles.scss
SunsetTechuila: When you use "import", sass does not add the contents of the imported file to the current file

.eslintrc.js
SunsetTechuila: But I can remove it if you want

theRealPadster (approved):

rxri (approved):

PR #656, opened 2023-11-19
fix: reset marketplace button

SunsetTechuila:
the reset button doesn't work right now

### DIFF
--- a/src/logic/Utils.ts
+++ b/src/logic/Utils.ts
@@ -213,17 +213,20 @@ export const generateSortOptions = (t: (key: string) => string) => {
     { key: "z-a", value: t("grid.sort.zToA") },
   ];
 };
+
+type MarketplaceResetCategory = "extensions" | "snippets" | "theme";
 /**
  * Reset Marketplace localStorage keys
  * @param categories The categories to reset. If none provided, reset everything.
  */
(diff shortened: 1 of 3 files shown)


theRealPadster:
This doesn't work? It worked originally.

SunsetTechuila:
Button click event is passed to the function as an argument rn

src/logic/Utils.ts
```diff
@@ -213,17 +213,19 @@ export const generateSortOptions = (t: (key: string) => string) => {
...
 /**
  * Reset Marketplace localStorage keys
  * @param categories The categories to reset. If none provided, reset everything.
  */
-export const resetMarketplace = (...categories: ("extensions" | "snippets" | "theme")[]) => {
+export const resetMarketplace = (categories?: ResetCategory[] | ResetCategory) => {
   console.debug("Resetting Marketplace");
 
   const keysToRemove: string[] = [];
+  const categoriesToReset = Array.isArray(categories) ? categories : (categories ? [categories] : []);
```
theRealPadster, line 225: Is there a reason why this was changed?

SunsetTechuila: i don't remember

theRealPadster (approved):

PR #661, opened 2023-11-30
chore(snippets): remove `Hide Upgrade Button`

SunsetTechuila:
> Hide Upgrade Button

adblock does this

> Better lyrics style

poorly made + beautiful lyrics extensions or bloom theme do the job better


theRealPadster (changes requested):
I can see the point in removing the "hide upgrade button" one, but the better lyrics one looks like it does work and provides a different UI for the lyrics, and at a lighter cost than running an extension

SunsetTechuila:
if you insist

theRealPadster (changes requested):
Oh, we can also remove the thumbnail image if we are removing the snippet.

theRealPadster (approved):

PR #671, opened 2023-12-09
feat: update react types to 18.2.0

theRealPadster:


SunsetTechuila:
react 18 is not fully backward compatible
and ofc all the new things are definitely not, since they are new

theRealPadster:
> react 18 is not fully backward compatible and ofc all of its new features definitely aren't because they are new

Don't think it matters? Newer versions of marketplace would be meant for newer versions of Spicetify/Spotify, so should only make sure the types are up to date.

rxri (approved):

PR #672, opened 2023-12-15
Resolve type errors: Quick Cleanup

CharlieS1103:


src/types/spicetify.d.ts
```diff
@@ -1997,10 +1997,12 @@ declare namespace Spicetify {
...
 	namespace Config {
 		const version: string;
-		const current_theme: string;
-		const color_scheme: string;
+		let current_theme: string;
+		let color_scheme: string;
 		const extensions: string[];
 		const custom_apps: string[];
+		let local_theme : string;
+		let local_color_scheme : string;
```
rxri, line 2005: why do you directly modify the spicetify types, don't do it please

SunsetTechuila: Or it is better to modify the types in the spicetify cli repo

rxri: Well, technically the `Config` object shouldn't be used for marketplace usage. It should be in its own object that only marketplace uses

SunsetTechuila: I think it was done this way to keep compatibility with extensions that check current theme and themes that check current color scheme

rxri: but they don't check the `local_theme` nor `local_color_scheme`, so this stuff could be kept in local storage for example

SunsetTechuila: Yes, `local_theme` and `local_color_scheme` definitely shouldn't be a part of the spicetify types declaration file

src/components/Modals/BackupModal/index.tsx
```diff
@@ -56,6 +56,7 @@ const BackupModal = () => {
    * Prompt user to select a file to import and then run importMarketplace
    */
   const importSettingsFromFile = async () => {
+  // @ts-expect-error: `showOpenFilePicker not defined in type Window`
```
theRealPadster, line 59: This one could probably be fixed somehow...

SunsetTechuila: `window["showOpenFilePicker"]()`

rxri: In this case, the window type should be extended, and definitely not resolved like @SunsetTechuila suggested

SunsetTechuila: Then https://www.npmjs.com/package/@types/wicg-file-system-access

theRealPadster: Yeah, if that package adds the type for it, we should probably add that 👍

theRealPadster (changes requested):
Please add that types package to resolve the `"showOpenFilePicker" not defined in type Window` error

rxri (changes requested):
This still should be done before merging, https://github.com/spicetify/spicetify-marketplace/pull/672#discussion_r1428481530

rxri:
No activity and none of requested changes have been implemented

PR #673, opened 2023-12-17
feat: don't install themes if `current_theme` isn't `marketplace`

SunsetTechuila:


CharlieS1103:
Is this something we really want?

SunsetTechuila:
Since themes don't work correctly without setting current_theme to `marketplace`, I think yes

SunsetTechuila:
I've seen a lot of people having issues with themes from the marketplace because they didn't do that

theRealPadster (approved):

PR #678, opened 2024-01-22
feat: localize notifications & app name

SunsetTechuila:


src/settings.json
```diff
@@ -1,5 +1,8 @@
 {
-  "displayName": "Marketplace",
+  "displayName": {
+    "en": "Marketplace",
+    "ru": "Маркетплейс"
+  },
```
theRealPadster, line 5: Is this a thing?

theRealPadster (approved):

src/settings.json
SunsetTechuila: https://github.com/spicetify/spicetify-cli/blob/86d5df56acd81cbdd72e3316d2821aa3045f61c8/CustomApps/lyrics-plus/manifest.json

theRealPadster: Huh, that's cool. Didn't realize.

PR #679, opened 2024-01-24
feat: option to show archived repos & `archived` tag

SunsetTechuila:


src/components/Card/TagsDiv.tsx
```diff
@@ -40,20 +41,25 @@ const TagsDiv = (props: {
...
   const extraTags = props.tags.slice(MAX_TAGS);
 
   // Render the tags list and add expand button if there are more tags
   return (
     <div className="marketplace-card__tags-container">
       <ul className="marketplace-card__tags">
-        { generateTags(baseTags) }
-        { extraTags.length && expanded
+        {generateTags(baseTags)}
+        {extraTags.length && expanded
```
theRealPadster, line 57:
```suggestion
        { generateTags(baseTags) }
        { extraTags.length && expanded
```

src/components/Card/TagsDiv.tsx
```diff
@@ -40,20 +41,25 @@ const TagsDiv = (props: {
...
           ? generateTags(extraTags)
           : null
         }
       </ul>
-      { extraTags.length && !expanded
+      {extraTags.length && !expanded
```
theRealPadster, line 62:
```suggestion
      { extraTags.length && !expanded
```

src/logic/FetchRemotes.ts
```diff
@@ -1,8 +1,6 @@
 import { BLACKLIST_URL, ITEMS_PER_REQUEST } from "../constants";
-import { CardItem, Snippet } from "../types/marketplace-types";
+import { RepoTopic, CardItem, Snippet  } from "../types/marketplace-types";
```
theRealPadster, line 2:
```suggestion
import { RepoTopic, CardItem, Snippet } from "../types/marketplace-types";
```

theRealPadster (changes requested):
Looks good to me. Just some linting stuff.

SunsetTechuila:
> some linting stuff

What linter should I use to see the warnings? Eslint doesn't show any. And every formatter removes those spaces

SunsetTechuila:
https://react.dev/learn#displaying-data

image

theRealPadster:
> you are showing a regular js, not jsx
It's in the `FetchRemotes.ts` file, like in the above comment/suggestion.

SunsetTechuila:
Yes, it is a regular js inside the jsx file

SunsetTechuila:
> like in the above comment/suggestion.

I have committed suggestion for the `FetchRemotes.ts`, I'm speaking about `TagsDiv.tsx` rn

theRealPadster:
> I have committed suggestion for the `FetchRemotes.ts`, I'm speaking about `TagsDiv.tsx` rn

Ah, maybe the ESLint rules don't apply to the tsx then, but I think it makes sense to follow the same conventions across both.

SunsetTechuila:
> I think it makes sense to follow the same conventions across both.

Are you sure? JSX has different conventions

theRealPadster:
It looks like [it's an option](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-curly-spacing.md). I'm sure there are different style guides, but I'm used to using spaces and regardless which we do, I think we should be consistent between js/jsx files in the same project.

SunsetTechuila:
image

image

theRealPadster:
> ew in jsx you shouldn't have curly spacing. looks bad af

Literally looks the exact same as in JS. 

But okay if it's all done without spaces currently then I guess let's leave it. 🤷

theRealPadster (approved):

rxri:
> > ew in jsx you shouldn't have curly spacing. looks bad af
> 
> Literally looks the exact same as in JS.
> 
> But okay if it's all done without spaces currently then I guess let's leave it. 🤷

tbh I have never seen JSX code with style like this. JS and JSX are not the same and shouldn't be connected with styles. JS has imports with curly spacing because that's how "it was designed" and JSX was designed to not have them

PR #701, opened 2024-03-19
feat: fetch snippets from github repo at runtime

theRealPadster:
I tested it locally by commenting out the fetch call and importing the `snippets.json` file, since this change isn't live yet.


rxri (approved):
ok

PR #729, opened 2024-04-07
fix(snippets): move previews to the correct directory

tsuliwaensis:
Moved preview pics to the `assets` subdirectory


theRealPadster (changes requested):
I think you need to update the snippet JSON file with the new paths if you change where they're located

tsuliwaensis:
> I think you need to update the snippet JSON file with the new paths if you change where they're located

The JSON file has the right path. It is the preview thumbnails that is in the wrong directory

rxri (approved):

PR #742, opened 2024-05-03
feat: add extension to the blacklist

rxri:
This extension uses `LyricsPlus` name and such is confusing to our users because they're greeted with modified Lyrics-Plus with some gradient background that cannot be disabled.
If the author will change the name, we can remove it from the blacklist

### DIFF
--- a/resources/blacklist.json
+++ b/resources/blacklist.json
@@ -1,3 +1,6 @@
 {
-	"repos": ["https://github.com/FoxRefire/spiceDL"]
+	"repos": [
+		"https://github.com/FoxRefire/spiceDL",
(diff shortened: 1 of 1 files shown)


Delusoire:
Yeah, the naming is definitely misleading.

Delusoire (approved):

rxri:
The extension doesn't add custom lyrics so the name is still *misleading*. It adds gradient background into the `lyrics-plus`, so you should choose the name accordingly and yea

dupitydumb:
well, im currently working to integrated the setting. but I get your point. i have renamed it to Lyrics Glow in the [manifest](https://github.com/dupitydumb/CustomLyrics-Spotify/blob/main/manifest.json)

PR #759, opened 2024-05-20
feat: switch to biome to ensure consistency & apply various fixes

afonsojramos:
This repo contains multiple files with different types of spacing, formatting done via eslint, which is not recommended by the team itself, so this pr introduces biome, which has already been implemented in [spicetify/cli](https://github.com/spicetify/cli).


rxri (approved):

rxri:
> Why are we using no trailing commas? Should be "always" for more readable diffs. 

Should be "es5". Didn't notice it. Not really much of the deal because the marketplace code is going to be replaced soon

PR #768, opened 2024-05-30
Update snippets.json

Plueres:
Removed my own redundant css, that doesn't work anymore because of the newer updates


rxri:
Marketplace works on versions 1.2.13 -> 1.2.38. Snippets should stay since they still can be used on older versions. You can add information to the description on what versions they work

PR #770, opened 2024-05-31
added 2 snippets regarding the global nav bar

Plueres:
1. removed the global nav bar item limit, so you don't have to scroll through the items
2. removed the forced margin inside the marketplace for certain themes with the global nav bar enabled


kyrie25 (changes requested):
Resolve conflict

resources/snippets.json
```diff
@@ -401,6 +401,18 @@
...
     "preview": "resources/assets/snippets/More-Visible-Unplayable-Tracks"
   },
+  {
+    "title": "[Global Nav Bar] remove scrolling overflow",
+    "description": "Removes the max visible width restrictions of the 'sidebar' items. Useful if you have more than the scrolling limit but not as much that it will overflow with the searchbar",
+    "code": ".custom-navlinks-scrollable_container{max-width: unset!important}",
+    "preview": "resources/assets/snippets/glob-nav-max-items.png"
+  },
+  {
+    "title": "[Global Nav Bar] marketplace header margin fix",
```
rxri, line 411: huh. this margin exists so the buttons are clickable. if you unset it, they won't be clickable again

Plueres: I have had certain themes make the menu overlap. though I have not disabled this in a long time so I'm not sure if that is still the case.

rxri (changes requested):
remove the image `glob-nav-mrktplc-margin.png`

resources/snippets.json
```diff
@@ -401,6 +401,12 @@
     "code": ".main-trackList-disabled{background:#f004}.main-trackList-disabled:focus-within,.main-trackList-disabled:hover{background:#f005}.main-trackList-disabled.main-trackList-selected,.main-trackList-disabled.main-trackList-selected:hover{background:#f006}",
     "preview": "resources/assets/snippets/More-Visible-Unplayable-Tracks"
   },
+  {
+    "title": "[Global Nav Bar] remove scrolling overflow",
+    "description": "Removes the max visible width restrictions of the 'sidebar' items. Useful if you have more than the scrolling limit but not as much that it will overflow with the searchbar",
```
kyrie25, line 406: Some wording suggestions because it sounds really confusing at first glance.
```suggestion
    "description": "Removes the max visible width restrictions of the 'sidebar' items. Useful if you have more than the scrolling limit but not so much that it will interfere with the search bar",
```

kyrie25 (changes requested):

PR #772, opened 2024-05-31
feat(i18n): add german translation

MikaHD666:


theRealPadster:
It looks like biome is really picky about imports and wants the "de" language imported after the "ca" one, to keep it alphabetical.

theRealPadster:
Now I think it's complaining about the extra empty line where the DE import used to be?

src/resources/locales/index.ts
```diff
@@ -12,6 +13,7 @@ import uk from "./uk.json";
 import zhCN from "./zh-CN.json";
 import zhTW from "./zh-TW.json";
 
+
```
theRealPadster, line 16:
```suggestion
```

theRealPadster (approved):

Issue #691, opened 2024-03-02
Hover panels snippet scrollbar issue


warpje5:
### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version: spotify-launcher 0.5.4-1 (latests from pacman official repository)
- Spicetify version: 2.33.2
```

### 📝 Description

Hi! When you install "Hover Panels" snippet and turn on "Now playing" panel, horizonal scrollbar appears at bottom of app. I think it begin appears around 2 week ago, so I need to keep "Now playing app" always closed to get rid of scrollbar.

### 📸 Screenshots

![2024-03-02T20:45:13,712000357+03:00](https://github.com/spicetify/spicetify-marketplace/assets/125624948/3ce3e818-7599-42da-b82d-413489e157fd)

Ender-Wang:
I am not using any theme and use Hover Panels without this issue. Better mention the theme dev about this.

Ender-Wang:
I made a new Hover Panel snippet and it's in PR ATM, I also noticed this today and also other style issues brought by the old `Hover Panels` snippets, so I made myself a better one lmao, but only focuses on the left side panel. #778

Ender-Wang:
Besides, let me know whether I should delete the old `Hover Panels` from two years ago since I am not the creator of that one.

PR #778, opened 2024-06-07
feat(snippets): add a ` New Hover Panel`

Ender-Wang:
Fix extra left margin in  snippet `Hover Panels`
hover-panels-fix-left-margin


Ender-Wang:
A New Hover Panel that servers the same purpose as the legacy 'Hover Panels', but only for the left sidebar and without modifying any other elements.
Issues related to the legacy `Hover Panels`:
- Messed up style with other components:
CleanShot_2024-06-08_12-11-44
- Miss behavior of the horizontal scrollbar: #691

Ender-Wang:
Besides, let me know whether I should delete the old Hover Panels from two years ago since I am not the creator of that one.

rxri (approved):

PR #784, opened 2024-06-16
feat: support user snippets in repositories

zhngharry:
Based on #435 

Implements the functionality that user hosted snippets in own repositories are fetched and shown in marketplace.

Important notes/alternatives considered:
- I've read through the discussion in #435, and believe it would be best to have the css code within the manifest, just like we are doing in the official repo right now
- If this PR is merged, the documentation needs to be adjusted; if a user wants to publicize their snippet from their own repo, they need to add the **spicetify-snippets** tag to their repo, as well as have a correct manifest entry with the following format:

image

Sidenote: PR #781 fixes the issue that users aren't able to host different types of creations within the same repo


rxri:
We have already discussed it on discord and I already said no to it. This is not something that should be implemented at EOL of v2 basically and it doesn't bring anything good. No one will use it regardless. I'm against of merging this pull request

theRealPadster (approved):
I think it's a good idea. It would mean people can add their own snippets without needing to submit a PR and have someone from the team take the time to review it. Snippets are very safe compared to extensions, which we allow without manual review.

theRealPadster:
What if we make it so it shows the built-in snippets from the repo but also adds support for user snippets? Otherwise we'd lose the ones we have unless we put them somewhere else.

zhngharry:
> What if we make it so it shows the built-in snippets from the repo but also adds support for user snippets? Otherwise we'd lose the ones we have unless we put them somewhere else.

Isnt this PR already implementing this, or are you talking abt v3? I don't know if we/the team wants user snippets in v3 because it would kinda disrupt the whole monomanifest flow, no?

To bypass the loss of the current snippets couldn't you just add them to the official modules repo then? Or would that also mean having to filter good/bad snippets for the official repo?

theRealPadster:
Ohh, my bad, I didn't read it through fully, I was thinking it just changed snippets to be user-based, which I think is what I had originally done.

CharlieS1103:
I think this all looks great! One thing is that we'd have to add support for authors just to ensure people are credited, additionally all the hrefs currently lead to the marketplace repo, we'd have to connect them to the correct repo. @zhngharry

rxri:
Inactivity. Also I noticed it doesn't even check the blacklist (or so i think). User snippets are bad idea.

PR #791, opened 2024-06-23
feat(snippets): add a `Dynamic Left Sidebar`

Ender-Wang:
####  Dynamic Left Sidebar Description
- Makes the left sidebar dynamic, so it only shows when you hover over it and pushes the main content to the right.
#### Motivation
- Previously, I made a `New Hover Panel` #778  that fixed the legacy `Hover Panels.` But after some time using it, I found it blocks the beginning of a playlist(can be anything) when I hover the left sidebar that activates the hovering, which is the expected hovering behavior, but it's kinda annoying. Then I made this one to be like when you hover the left side edge, the LeftSidebar appears and pushes the content to the right so nothing is blocked.


theRealPadster (approved):

theRealPadster:
Does this replace the other new hover panel snippet? Should we remove it in the code?

Ender-Wang:
> Does this replace the other new hover panel snippet? Should we remove it in the code?

Function-wise, it replaces all hovering left sidebar snippets, both can be there in case someone prefers hovering instead of this dynamic one.

theRealPadster:
Oh, the hover one goes on top of the content vs this moves it to the side?

Ender-Wang:
Exactly. This one pushes the main view to the right without blocking anything below it.

theRealPadster (approved):

PR #804, opened 2024-07-18
feat (snippet): Implement a tweak to adjust Now Playing alignment

moazie:
This tweak allows for a user to adjust the positioning of the Now Playing widget from the left.
Now-Playing-Alignment-Tweak


rxri:
That's an issue on your end though, either by theme or snippet. The second screenshot is the stock "setting"

rxri:
also your snippet doesn't allow to change anything. It assigns the value and user CANNOT modify it.

moazie:
> That's an issue on your end though, either by theme or snippet. The second screenshot is the stock "setting"

Thats true, i was using another snippet which moves the sidebar to the left so it looked a bit off. This is also my first pull request so i dont have that much experience as you can see.

PR #810, opened 2024-07-27
feat(snippets): adding oneko for the comfy theme

BeBeepBOOPbot:
this is the oneko snippet for the [Comfy theme](https://github.com/Comfy-Themes/Spicetify)
this is almost the same as the og Oneko snippet but i changed its position and i bumped up the z index so the cat is on top of the bar
this is my first pull request so tell me if i messed anything up
and if its too similar to the og Oneko snippet then this pull request can be disregarded


rxri (changes requested):
edit the snippet to work on comfy as well, and do not create new one just for this purpose

PR #829, opened 2024-08-31
fix(snippets): sidebar switch w/ no now playing sidebar

zoetrope69:
When now playing sidebar is collapsed it was breaking because it was taking up the wrong amount of space


kyrie25 (commented):
Is the change still relevant/working?

rxri:
Inactivity. Closing

PR #836, opened 2024-09-15
fix(snippet): update scrollbar snippets

sanoojes:
update scrollbar snippet stylesheet


kyrie25 (commented):
Is the change still relevant/working as expected?

kyrie25 (approved):

PR #849, opened 2024-10-07
fix(snippet): pointers

Climax12:
Added pointer to main-card-card on marketplace, main track List Rows and sort by objects


theRealPadster (approved):

rxri:
seriously, with the space?

rxri:
GitHub follows the first pull request title when merging. Get refined github that fixes it or smth @theRealPadster

theRealPadster:
Ah sorry, I never remember. Why does it use the first title and not what's currently there? That's so dumb. Also there is no refined GitHub on my phone.

Climax12:
> Ah sorry, I never remember. Why does it use the first title and not what's currently there? That's so dumb. Also there is no refined GitHub on my phone.

sry i mistakenly place a space :)
I think GitHub follows the Commit title when merging.

rxri:
> Ah sorry, I never remember. Why does it use the first title and not what's currently there? That's so dumb. Also there is no refined GitHub on my phone. 

Idk why it exists but it does

PR #852, opened 2024-10-18
feat(i18n): add pt-BR support

Kyo-70:
Here's my contribution to adding my Brazilian Portuguese language to the program.


theRealPadster:
Thanks for the translation. Before it can be used, it needs to be registered in the locale system. Could you do this step here as well? 
https://github.com/spicetify/marketplace/wiki/Localizing-Marketplace#registering-your-language

theRealPadster (changes requested):
The file also needs a .json extension

Kyo-70:
> Thanks for the translation. Before it can be used, it needs to be registered in the locale system. Could you do this step here as well? https://github.com/spicetify/marketplace/wiki/Localizing-Marketplace#registering-your-language

I've just made the change and I hope it's correct.

Kyo-70:
I'll apologize because I'm not able to integrate another file for change I'm new to the github tools... if you can help me how it's done or just include and edit I'd appreciate it

Kyo-70:
Now I've done it - update file index.ts

src/resources/locales/index.ts
```diff
@@ -12,6 +12,7 @@ import ru from "./ru.json";
 import uk from "./uk.json";
 import zhCN from "./zh-CN.json";
 import zhTW from "./zh-TW.json";
+import ptBR from "./pt-BR.json";
```
theRealPadster, line 15: It looks like the linter is complaining that these aren't in alphabetical order 🙄 
Could you put the ptBr one in between "pl" and "ru"?

theRealPadster (changes requested):

theRealPadster (approved):
Passed the checks finally, thank you!

PR #853, opened 2024-10-19
update index.ts add PT-BR

Kyo-70:
I've added it to the index.ts file and I hope it's right


rxri:
Do it in #852

Kyo-70:
> Do it in #852

I'm sorry I did it wrong, I'll put it in the original post

PR #861, opened 2024-11-18
new_snippet_slide_in_sidebar

Aayush-K15:
A new snippet has been added. There is a similar one but that messed with the now playing bar margin in the bottom also this has smoother transition and a snappier one. Better for the users to have an option. Also my first contribution to Spicetify I've loved it and been using it for so long, all of my friends have it as well, we all love it and its my honour to be able to contribute to such a project. KUDOS TO YOU.


kyrie25 (commented):
Is the snippet still working as of `1.2.55`?

PR #866, opened 2024-11-30
feat: add Hide Update Spicetify Button

suiminn:


theRealPadster:
What is the point of this? Shouldn't people just update their Spicetify if the button shows up?

theRealPadster:
I'm not sure how I feel about making it easy to hide the notification that your Spicetify is out of date. I can see how it would be helpful if you're intentionally using an older version and want to lock things as they are, but for most people, I don't think they'd need this and making it easy to do could result in new users hiding the notification if they don't know how to update or are frustrated if it shows up too often. If you're locking versions manually you're probably advanced enough to be able to add this as a custom snippet.

rxri:
Shouldn't be available on marketplace.

PR #892, opened 2025-02-08
fix(snippets): smaller right sidebar cover art

Climax12:
image


Climax12:
and fixed `Smaller right sidebar covert art` -> `Smaller right sidebar cover art`

kyrie25 (approved):

PR #906, opened 2025-02-21
feat(snippets): add hide create button

Climax12:
Added snippet to hide Create button.

Hide-Create-button


rxri:
create button? where even is that?

rxri:
ah i see now, but you can disable it with exp feature so no need for snippet
image

Issue #887, opened 2025-01-31
the navbar overlaps when performing a search with no results


Botato300:
### 🔍 Is there already an issue for your problem?

- [x] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version:
Spotify for Windows (64 bit)
1.2.56.502.ga68d2d4f
Spicetify v2.38.7
xpui_2025-01-28_1738065213461_a68d2d4
cef_131.3.1+gcb062df+chromium-131.0.6778.109
Runtime: Chrome
Build System: Bazel

- Spicetify version: 2.38.7
```

### 📝 Description

When you perform a search from the Marketplace and get no results, the navigation bar overlaps with the search bar, and does not let you interact.

![Image](https://github.com/user-attachments/assets/04ec8194-7caf-4cd5-b987-ca09958aabd5)

Interestingly, in the 'Extensions' and 'Snippets' tabs, this does not happen. It works correctly. The problem only happens in the tabs: Themes, Apps, Installed.

![Image](https://github.com/user-attachments/assets/a9cead92-4cc2-4e1d-bc6b-f0837c3ebf15)

It could be solved by adding CSS properties margin-top, or padding-top to the container but I don't know if that's the way to go.

S1NJED:
i have the same issue

PR #913, opened 2025-03-01
fix: icons and "Sort By" div overlapping

S1NJED:
Related issue: #887

The issue: The `LoadMore` and `LoadIcon` icons were rendered behind  the "SortBy" div (which is a sticky div) and the sortBy div appeared behind the first div when the search do not return elements.

Before: 

https://github.com/user-attachments/assets/4aafa21f-62c0-472d-a7a3-7200f622fb82

After:  

https://github.com/user-attachments/assets/5f2b11df-9b8f-4ac5-9dbe-9e19875c3fc7

For the Snippets tab there is still the "+ Add CSS" button overlapping but I couldnt figure out how to fix this.

### DIFF
--- a/src/components/Grid.tsx
+++ b/src/components/Grid.tsx
@@ -643,7 +643,7 @@ class Grid extends React.Component<
           if (cardsOfType.length) {
             return (
               // Add a header for the card type
-              <>
+              <div className="marketplace-content">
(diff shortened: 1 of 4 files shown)


rxri (approved):
that's cool, thanks

PR #920, opened 2025-03-14
 feat(snippets): add now playing never dims snippet 

DarkBlu242:
keeps the now playing view from dimming as if it's always being hovered


rxri (changes requested):
the class should be mapped in the cli,
we do not allow external services for previews. upload it to the `snippets` directory like any other snippet.

PR #937, opened 2025-04-10
feat(snippet): add "hide 'scroll through previews'"

Donaldino7712:


rxri (changes requested):
This element should be mapped inside main repo (cli) to some name that would be the most fitting one and then we can merge this pull request after changing the classname to mapped one

rxri (approved):

PR #940, opened 2025-04-14
feat(snippets):CSS snippet to increase the width of the apps list at the top left of spicetify

Flanpit:
With more than 4 apps installed, a few get cut off. This just increases the width of that area.


rxri:
This limit (4) exists for a reason - so when you have small window, the buttons won't overlap. If you want, keep it as custom snippet, locally, but this won't be merged

PR #941, opened 2025-04-23
feat(snippets): star wars progress bar

1-MaX-7:
light-saber-bar


rxri:
This is already built-in to spotify though. Just enable exp feature for it iirc

PR #953, opened 2025-05-08
feat(snippet): add "Fix queue popup colors"

Firepup6500:


rxri:
All modals are in light mode. To be honest, it should be up to theme to change it.

PR #954, opened 2025-05-08
feat(resources/snippets): add hover panel (tm)

bdsqqq:
hey, let me know if a preview image is a must, if so I'll add one 🫶

### DIFF
--- a/resources/snippets.json
+++ b/resources/snippets.json
@@ -1,5 +1,10 @@
 [
   {
+    "title": "hover panel(tm)",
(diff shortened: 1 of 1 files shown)


rxri:
Snippet for hovering left panel is already added into marketplace. It's called `New Hover Panels` or `Hover Panels`
https://github.com/spicetify/marketplace/blob/main/resources/snippets.json#L75 https://github.com/spicetify/marketplace/blob/main/resources/snippets.json#L81

PR #959, opened 2025-05-15
feat: add a Sonic Dancing that moves with the progress bar snippet

phuid:
I know a sonic dancing snippet already exists, but i believe it really could use the functionality of moving with the progress bar. I didn't feel authorized to delete the original snippet tho.


rxri:
Either modify the original snippet or just use https://github.com/BlafKing/spicetify-cat-jam-synced

PR #972, opened 2025-06-11
feat(i18n): add finnish translation, fix a typo

Venni1:
There was a typo in a key "almbumArtColorsModeToolTip", so I fixed it. Also added the Finnish translation.

PR #975, opened 2025-06-12
feat(i18n): add finnish translation

Venni1:
Added the Finnish translation. Split from #972.


theRealPadster (changes requested):
Needs Finnish translation added to app.tsx

theRealPadster (approved):

PR #991, opened 2025-07-07
feat(settings): export as file & upgrade deps

kyrie25:


src/components/Modals/Settings/DnDList.tsx
```diff
@@ -41,7 +19,7 @@ const DnDList = (props: { modalConfig: Config; updateConfig: (CONFIG: Config) =>
       justifyContent: "center",
       textDecoration: isEnabled ? "none" : "line-through",
       cursor: "pointer",
-      ...style
+      ...draggableStyle
```
rxri, line 22: It's fixed though? I added this stupid workaround because the items were going pretty far off from the cursor

kyrie25: Seems the new library fixed it because this workaround made the draggables fly off screen when I drag them lol

rxri (approved):

PR #1001, opened 2025-08-06
feat(snippets): an addition to snippets for adding a moving pointing man GIF to the progress bar

kebiBGit:
This is inspired by previous snippets and is meant to enhance the user interface by providing a visual cue that follows the playback progress.


kyrie25 (changes requested):
Preview image should be taken of the client.

PR #1008, opened 2025-08-30
feat(snippets): Gawr Gura Karaoke

PiggMelonie:
Added a gawr gura gif that goes over the progress bar


rxri:
coderabbit did not request anything from you to fix so i don't understand why you closed this. the `preview` image should be taken from the spotify client

PR #1093, opened 2026-01-17
feat(snippets): add Hide Lyrics Button snippet

pnxl:
This implementation differs from the existing one on the snippets library as it targets the button with the aria-label tag, instead of the whole class.

Targeting the whole class and adding `display: none` also hides the buttons around the lyrics button.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the “Hide lyrics button” snippet to consistently hide the lyrics control in the player.
  * Improved targeting so other player controls remain unaffected.


pnxl:
I'd like to clarify that what the AI is saying isn't true lol, the other hide lyrics snippet is targetting with the CSS class `.main-nowPlayingBar-lyricsButton`, which when set to `display: none` hides the buttons around it too:
CleanShot 2026-01-17 at 22 36 03

This snippet fixes that by targetting the lyrics button with its aria-label and only hides the Spotify lyrics button:
CleanShot 2026-01-17 at 22 36 41

rxri:
The AI is right; this is a duplicate entry. You should update the existing snippet and not add another under the same name. Also, targeting via aria-label makes this css only work on English language.
```css
.main-nowPlayingBar-lyricsButton[data-testid="lyrics-button"] { display: none !important; }
```
This is more valid way to hide it on new versions

pnxl:
Smart, I'll add that to my fork. Also, I didn't know you're allowed to update an existing snippet, haha - sorry!

rxri (approved):

PR #1109, opened 2026-02-24
feat: add another downloader

rxri:
## Summary by CodeRabbit

* **Chores**
  * Updated configuration data with new repository entries and pattern allowances to maintain current data standards.

### DIFF
--- a/resources/blacklist.json
+++ b/resources/blacklist.json
@@ -7,6 +7,7 @@
     "https://github.com/ssatwik975/Speedify",
     "https://github.com/unknownkneega/complex-lyrics",
     "https://github.com/UmairAhmed406/spicetify-spotiflac",
+    "https://github.com/diorhc/Spicetify-Downloader",
     "// for new bl syntax:",
     "https://github.com/FoxRefire/*",
     "https://github.com/*/spiceDL",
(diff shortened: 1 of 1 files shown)


aazz180:
> Downloading music files locally to use outside of Spotify is not even a premium feature, it's something you're not allowed to do period.

Okay, but why do you disallow via filtering from anyone that is using the Spicetify Marketplace, isn't that Github/Spotify's lawyers job, it doesn't affect this repo if you don't add it to the marketplace. I just don't understand the logic behind this PR. 

Aren't you using more of the user's resources to check and block it, and the marketplace is opensource, so it's easy to bypass.

theRealPadster:
> Okay, but why do you disallow via filtering from anyone that is using the Spicetify Marketplace, isn't that Github/Spotify's lawyers job, it doesn't affect this repo if you don't add it to the marketplace. I just don't understand the logic.

Because we don't want to give them any reason/excuse to take down the marketplace or Spicetify. If someone uploads a downloader that makes use of Spicetify, that's one thing, but we would rather not have it show up for easy install in the marketplace, so it doesn't look like we're facilitating or condoning piracy.

aazz180:
> > Okay, but why do you disallow via filtering from anyone that is using the Spicetify Marketplace, isn't that Github/Spotify's lawyers job, it doesn't affect this repo if you don't add it to the marketplace. I just don't understand the logic.
> 
> Because we don't want to give them any reason to take down the marketplace or Spicetify. If someone uploads a downloader that makes the use of Spicetify, that's one thing, but we would rather not have it show up for easy install in the marketplace.

But that's not what this PR is doing, this isn't adding it to the marketplace - this downloader just makes uses of spicetify and this PR isn't adding its functionality to the market place. Your adding it to a blacklist of plugins that is disallowed? I understand malicious plugins, but why this.

PR #1117, opened 2026-03-06
feat(snippet): add enlarge_app-icons

xeraox3335:
Adds a new snippet to the `resources/snippets.json` file that allows users to enlarge custom app icons by expanding the relevant UI sections, making them easier to view without scrolling.

New snippet added:
* Added an "Enlarge App Icons" snippet that updates CSS to expand the custom app icon section and related UI areas for better visibility and usability.
enlarge_app-icons

## Summary by CodeRabbit

* **New Features**
  * Added a new snippets library entry titled "Enlarge App Icons" — a ready-to-use styling snippet with a descriptive overview and a visual preview to help users quickly apply larger app icon layouts.

PR #1140, opened 2026-04-26
feat(snippets): added a snippet to show all buttons on top bar

FoxyLoon:
Expand the button bar at the top to its full width and adjust the search bar

## Summary by CodeRabbit

* **New Features**
  * Added a new CSS customization preset for navigation styling, enabling improved layout control for navigation links and search positioning.


rxri:
Duplicate of https://github.com/spicetify/marketplace/pull/1117
I also explained why we don't want this merged

FoxyLoon:
> Duplicate of #1117
> I also explained why we don't want this merged

@rxri It might be intentional... but not everyone likes having those buttons hidden. The “carousel” with the apps also moves too fast to see the other apps, which makes the “user experience” annoying.

Giving users the option or choice to show everything or not, rather than blocking them in a specific way, would be a better choice in my opinion, especially when the app tends to offer endless customization.

This is just my opinion. The final decision is yours. Keep up the good work. 👍

PR #1165, opened 2026-06-20
fix(snippets): update hashed classes for gigantic npv canvas

Mythicsoul:
Updated classes and adjusted the "Switch to video" buttons position.
Creating as a draft since some of the classes are not mapped yet, so I'll go ahead and do that first.


rxri:
stale

PR #1204, opened 2026-07-29
feat(snippets): add Hide Playlist Ads snippet

h3j4:
Hides sponsored content in playlist and Now Playing headers: the video ad, the "Limited ads • Presented by X" row in both places, and the sponsor card section.

Every selector anchors on data-testid, href, or css-map class names rather than Spotify's hashed classes. Each target has two independent anchors so a single rename doesn't disable the rule.

## Summary by CodeRabbit

* **New Features**
  * Added a “Hide Playlist Ads” snippet to remove sponsored playlist video ads, ad labels/overlays, and sponsor-card sections from the page.  
  * Included a preview asset for the new snippet.


h3j4:
Good catches, both addressed.

1. Confirmed: a playlist description containing an external link matched the old selector. Re-anchored on the ad-server host (servedby. / flashtalking) so only sponsored rows match.

2. Added section:has([data-encore-id="listRow"] a[target="_blank"][href^="http"]) as a locale-independent anchor. data-encore-id alone would be far too broad, but combined with an off-site link it's ad-specific: Spotify's own list rows link to internal paths. Verified against a localized card with an unrecognised ad host.

On localized fixtures: snippets in this repo are CSS strings in snippets.json with no test harness, so there's nowhere to add them. Happy to include the markup I tested against in the PR description if that's useful.

rxri:
Marketplace snippets are not a place to do ad-hiding functionality

PR #1212, opened 2026-08-04
fix: await install persistence before reload

### LINKED ISSUE #1186: Remove extension
### 🔍 Is there already an issue for your problem?

- [x] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version: 1.2.93.667.g7b5cc0ce
- Spicetify version: Spicetify v2.44.0
```

### 📝 Description

after attempting to remove an extension (in my case shuffle+) and reloading, extension is not functional but still shows up as installed in the marketplace, and it doesnt show up in the installed section.

also reloading hides spicetify so i have to `spicetify apply` for it to show up again

### 📸 Screenshots

Image

yassinsolim:
## Summary
- use the awaitable storage methods for extension and theme lifecycle changes
- serialize mutations with Dexie and keep related writes atomic
- hydrate storage inside the mutation queue before snapshotting durable state
- persist both the item and installed-item list before showing the reload modal
- await theme CSS apply/removal and the same writes during automatic updates
- surface async lifecycle failures through existing error notifications

## Root cause
The synchronous storage wrappers updated their in-memory cache while starting IndexedDB transactions without awaiting completion. Reloading immediately could therefore race the durable install or removal state.

There was also a startup race: a user could remove an extension while IndexedDB hydration was still in flight. Late hydration then replaced the cache with the old durable snapshot, causing the queued mutation to persist stale state and make the extension reappear after reload. Mutations now await hydration inside the serialized queue before cloning their previous/next states.

## Validation
- `pnpm install --frozen-lockfile`
- `pnpm biome check src/components/Card/Card.tsx src/logic/Storage.ts src/logic/Utils.ts`
- `pnpm build:local`
- targeted behavioral harnesses for serialized mutations, rollback after failed persistence, no-IndexedDB fallback, delayed-hydration uninstall, and durable removal through a fresh IndexedDB connection
- live Spotify 1.2.95.453 on macOS with Spicetify 2.44.0: removed Soundalike, accepted reload, confirmed Installed remained empty after reload, then reinstalled it and confirmed the delete affordance returned

The repository's existing TypeScript baseline still reports the same six unrelated errors; this branch adds no new TypeScript diagnostics.

Closes #1186


src/components/Card/Card.tsx
```diff
@@ -114,32 +114,33 @@ export class Card extends React.Component<
         console.debug(`New update pushed at: ${pushed_at}`);
         switch (this.props.type) {
           case "extension":
-            this.installExtension();
+            await this.installExtension();
             break;
           case "theme":
-            this.installTheme(true);
+            await this.installTheme(true);
             break;
```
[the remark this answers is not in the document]
yassinsolim: Fixed in 1e3a2df: theme install/removal now await the CSS chain, CSS errors propagate, and the button and mount entry points log failures and show the existing type-specific error notification.

theRealPadster (dismissed):
Can you address the Copilot comments please?

theRealPadster:
Thanks for the iterations here. Rather than send you round again I've pushed four commits to your branch — a couple of them touch files outside the original diff.

**`fix(themes): reject non-OK theme CSS responses`** — this is CodeRabbit's outstanding point. `parseCSS()` read `res.text()` without checking `response.ok`, so a 404/500 body would be injected as the theme's CSS and the theme still persisted as installed. Worth noting the knock-on: `extension.tsx` calls `initializeTheme()` at startup without awaiting or catching it, so making `parseCSS` throw would have skipped the theme's `include` scripts on a CSS failure. I guarded the CSS step there so the JS still loads.

**`fix(storage): keep hydrating when IndexedDB is unreadable`** — this was the actual blocker for me. `runStoreRequest()` now throws instead of resolving `undefined` when the database is unavailable, which is correct for writes, but it meant a failed `loadIndexedDBCache()` aborted `hydrateMarketplaceStorage()` before `migrateLocalStorage()` ever ran. On a client without IndexedDB, or with a blocked/corrupt database, the cache stayed empty and every installed extension and theme showed as uninstalled — previously we degraded to an in-memory cache populated from localStorage. Only the read path is tolerant now; writes still reject so failures surface, which is the point of this PR.

**`fix(storage): don't evict cache keys a mutation never owned`** — `mutateStorage()` reconciles by deleting every cached key absent from the mutation result, but that result derives from `durableCache`. `migrateLocalStorage()` sets `cache` first and `durableCache` only per successful `persistItem()`, so a partial migration leaves the two out of sync and the next successful mutation silently drops the un-migrated keys. Deletion is now limited to keys present in the durable snapshot the mutation started from.

**`fix(card): salvage valid entries from a malformed installed list`** — `readStoredStringArray()` returned `[]` if any single entry wasn't a string, and callers write the result straight back, so one bad entry in `installedExtensions`/`installedThemes` turned into a wiped list on the next install or removal.

Two things left for you:

- `removeTheme()` (Card.tsx:444) is now unreferenced — `toggleTheme()` calls `removeThemeNow()` directly. Worth dropping.
- The new `schemesURL` `.ok` check in `prepareTheme()` is stricter than before: a theme whose schemes URL is momentarily unavailable now fails to install outright, where previously it installed scheme-less. Happy either way, just flagging it as a deliberate call.

Typecheck is unchanged against main (same two pre-existing errors) and biome is clean on the touched files.

src/components/Card/Card.tsx
```diff
@@ -272,54 +300,56 @@ export class Card extends React.Component<
...
+      lastUpdated,
+      created
+    });
 
-    // TODO: handle this differently?
+    let userCSS: string | undefined;
+    if (!item.include) {
+      const tld = window.sessionStorage.getItem("marketplace-request-tld") || undefined;
+      userCSS = await parseCSS(item, tld);
+    }
```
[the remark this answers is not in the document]
theRealPadster: Fixed in 6fb5866 — `parseCSS()` now throws on a non-OK response. Note `extension.tsx` calls `initializeTheme()` unawaited at startup, so the CSS step there is wrapped to keep the theme's `include` scripts loading when CSS fetching fails.

rxri:
why you all complicate this that much. just use dexie at this point

theRealPadster:
> why you all complicate this that much. just use dexie at this point

Never heard of dexie. I just had Claude take a look at what was left since it's been a while and I couldn't remember what exactly was left or if it's okay to merge.

rxri:
https://dexie.org - wrapper around IndexedDB. Would cut 80% of the added code right now

yassinsolim:
Hey @rxri I refactored the persistence layer to Dexie in 92e5dfb, cutting the custom IndexedDB machinery substantially while keeping the existing synchronous cache API and ordered/atomic async mutations. I also moved snippet record/list updates into one awaited mutation, fixed the uppercase vibrancy regression, and removed the now-unused theme wrapper in 934856d. 

Validation: touched-file Biome check and the production build pass. Targeted harnesses passed for queued IDB mutations, no-IDB localStorage hydration plus async set/remove, forced Dexie write rejection with optimistic-cache rollback, async rejection, and queue recovery. TypeScript reports the same six pre-existing diagnostics as current main; there are no new diagnostics from this refactor. Could you take another look when you have a chance?

rxri:
I would appreciate if we would cut AI from our conversations because it's pretty obvious. If you don't understand the code AI writes or you don't have "enough time" and use tools to automatize everything you do, then the code is not worthy being merged. We're humans, not computers, so let's talk human-to-human. Regarding the pull request, I will check it in my own free time

yassinsolim:
Fair enough, I’ll keep the PR discussion more direct and in my own words going forward. I only use it to ensure clarity when discussing changes so I do not forget to mention something and cause issues with understanding moving forward.

For clarity, I do use AI-assisted tooling during development, but I review the changes, understand what’s being changed, and I’m responsible for the code I submit. I wouldn’t expect something to be merged solely because a tool generated or validated it.

src/logic/Storage.ts
```diff
@@ -1,141 +1,190 @@
...
+  for (const { key, value } of records) {
+    cache.set(key, value);
+    persistedCache.set(key, value);
   }
 }
 
+// A read that fails means the stored state is unknown, not that it is empty.
+// Reopening usually works, since a reload can race the previous connection
+// being torn down. If every attempt fails we give up rather than report an
+// empty Marketplace and let the next write overwrite what we could not read.
```
rxri: unnecessary comment

```suggestion
```

src/logic/Utils.ts
```diff
@@ -531,7 +531,10 @@ export const parseCSS = async (themeData: CardItem, defaultTld?: string) => {
   const assetsUrl = userCssUrl.replace("/user.css", "/assets/");
 
   console.debug("Parsing CSS: ", userCssUrl);
-  let css = await fetch(`${userCssUrl}?time=${Date.now()}`).then((res) => res.text());
+  const response = await fetch(`${userCssUrl}?time=${Date.now()}`);
+  // Otherwise an error page body gets injected as the theme's CSS
```
rxri: as above
```suggestion
```

src/app.tsx
```diff
@@ -122,7 +124,16 @@ class App extends React.Component<
   }
 
   async componentDidMount() {
-    await hydrateMarketplaceStorage();
+    try {
+      await hydrateMarketplaceStorage();
+    } catch (error) {
+      // Showing the grid would claim nothing is installed, and the next action
+      // the user takes would then save that empty state over the real one.
```
rxri: as above
```suggestion
```

src/components/Card/Card.tsx
```diff
@@ -331,35 +361,71 @@ export class Card extends React.Component<
       if (name) Spicetify.Config.current_theme = name;
       // @ts-expect-error: Cannot assign to 'color_scheme' because it is a read-only property
       if (activeScheme) Spicetify.Config.color_scheme = activeScheme;
+    } else if (previousThemeKey) {
```
rxri:
```suggestion
    } else if (previousThemeKey && previousThemeKey !== this.localStorageKey) {
```

rxri (changes requested):

rxri (approved):

PR #1218, opened 2026-08-12
fix: show a message when the installed tab is empty

yassinsolim:
## Summary

The Installed tab renders a blank page when you have nothing installed. There is already a "No installed X" message in the CSS, but it can't show up in that situation, so this adds a real message instead.

## Why it never showed

`_card.scss` puts the message on the grid container with `:empty::after`:

```scss
&[data-tab='Installed'] {
  &:empty::after {
    content: 'No installed ' attr(data-card-type);
  }
}
```

`Grid.tsx` only renders that container when the category has at least one card:

```tsx
if (cardsOfType.length) {
  return ( ...heading and grid... );
}
return null;
```

So with nothing installed there is no container, and nothing to attach the message to. The only time it does appear is right after you uninstall something without reloading, because the card is still in the list but `Card.render` returns null, which leaves the container in the DOM with no children.

Both of these came in with #396, so it has been this way for a while.

## Changes

- Move the per card type sections into a variable in `Grid.tsx` so the render can tell when every section is empty, and show a message on the Installed tab when that happens.
- Separate message when a search on the Installed tab matches nothing, which was also blank before.
- Remove the `:empty::after` rule and the `data-card-type` attribute it used. Nothing else references that attribute.
- The two new strings go through i18n. The old one was hardcoded English in CSS.

## Notes

The refactor in `Grid.tsx` looks bigger than it is. The card filtering and mapping code is unchanged, it just moved above the return so its result can be checked.

## Summary by CodeRabbit

* **New Features**
  * Added clear empty-state messages for installed items when none are available or when search results have no matches.
  * Improved the installed-items layout with centered, readable empty-state styling.

* **Bug Fixes**
  * Prevented duplicate cards and empty card sections from appearing in the marketplace grid.
  * Removed outdated empty-state messaging for individual card types.
  * Improved card organization and display consistency across marketplace sections.


src/components/Grid.tsx
```diff
@@ -534,6 +534,61 @@ class Grid extends React.Component<
...
+          return React.cloneElement(card, {
+            activeThemeKey: this.state.activeThemeKey,
+            key: card.key
+          });
+        })
+        .filter(
+          (
+            card,
+            index,
+            cards // Filter out duplicates to prevent spamming
```
rxri:
```suggestion
            cards
```

src/components/Grid.tsx
```diff
@@ -534,6 +534,61 @@ class Grid extends React.Component<
...
+          <div className="marketplace-grid main-gridContainer-gridContainer main-gridContainer-fixedWidth" data-tab={this.CONFIG.activeTab}>
+            {cardsOfType}
+          </div>
+        </div>
+      );
+    });
+
+    // The installed tab has nothing else to fall back on, so say so rather than
+    // leaving the user looking at an empty page. Wait for the load to finish
+    // first, otherwise this shows up for a moment on every visit to the tab.
```
rxri:
```suggestion
```

src/components/Grid.tsx
```diff
@@ -534,6 +534,61 @@ class Grid extends React.Component<
 
   render() {
     const { t } = this.props;
+
+    // Each card type renders its own heading and grid, or nothing when it has
+    // no cards to show.
```
rxri:
```suggestion
```

src/components/Grid.tsx
```diff
@@ -534,6 +534,61 @@ class Grid extends React.Component<
...
+          return (
+            !searchValue ||
+            title.toLowerCase().includes(searchValue) ||
+            user?.toLowerCase().includes(searchValue) ||
+            authors?.some((author) => author.name.toLowerCase().includes(searchValue)) ||
+            [...(tags ?? [])]?.some((tag) => tag.toLowerCase().includes(searchValue))
+          );
+        })
+        .map((card) => {
+          // Clone the cards and update the prop to trigger re-render
```
rxri:
```suggestion
```

rxri (changes requested):

rxri (approved):

PR #1209, opened 2026-08-03
fix: validate marketplace manifests with zod

yassinsolim:
## Summary

- validate cached and fetched remote manifests once at the input boundary with Zod
- keep required identity fields strict while normalizing optional previews, authors, tags, paths, and arrays
- parse multi-manifest repositories entry by entry so one invalid entry does not hide valid siblings
- preserve unknown manifest fields used by installation logic

This replaces #1208 and follows the maintainer feedback there: validation is centralized in `FetchRemotes.ts`; `Grid`, `Card`, and `Utils` are unchanged.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm build`
- `pnpm exec biome check --line-ending=crlf src/logic/FetchRemotes.ts`
- `git diff --check`
- `pnpm type-check` remains at the same six pre-existing diagnostics as `origin/main`
- live Spotify 1.2.94 / Spicetify 2.44 test loaded every extension page (153 -> 331 -> 367 cards), then searched for Soundalike successfully with no Marketplace React error boundary or runtime exception

Fixes #1203
Fixes #1207

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository manifest handling by validating entries and normalizing optional details.
  * Invalid or incomplete manifest entries are now excluded instead of causing processing issues.
  * Corrupted cached manifest data is automatically removed and fetched again.
  * Validated results are cached more consistently for faster, more reliable access.

Issue #1207, opened 2026-08-03
Extensions grid crashes on load-more: broken manifests on pages 2-4 crash the whole grid


LuiisDev21:
**Describe the bug**
Clicking "Load more" (or scrolling to the bottom) in the Extensions tab crashes the Marketplace with Spotify's "Ha habido un error / There has been an error" error screen. The crash makes pages 2-4 of the extensions grid unreachable.

**To Reproduce**
1. Open Marketplace → Extensions (topic search currently returns ~347 repos, 4 pages of 100).
2. Scroll to the bottom / click the load-more icon.
3. At some point during page 2-4 loading, the whole grid crashes into the error boundary.

**Root cause evidence**
I scanned the manifests of all repos on pages 2-4 (the exact query the grid uses: `api.github.com/search/repositories?q=topic:spicetify-extensions&per_page=100&page=N`). **40+ repos have missing or invalid manifests**:

- Invalid JSON (6): `NeonLightning/localSearch`, `donerschnitzel/SpicetifyCustomCover`, `EllisG17/immersive-player`, `GodKingTyler/guitar-hero`, `HardyUA/spicetify-extension`, and others
- HTTP 404 (36+): `nmattis/spicetify-extensions`, `Hyperiya/Iris`, `aiden04/spicetify-gui`, `cj123/Sensibletify`, `SplashFix/spicetify-lyrics-plus`, `DarkVanityOfLight/RustyLyr`, `ahmadzip/lyricfloat`, `jamesrchen/Spicetify-volumePercent`, etc.

`fetchRepoManifest` handles fetch/parse failures defensively (worker error → null → "noManifests"), so the crash is downstream: either `fetchExtensionManifest`'s reduce or the Card render throws on a malformed manifest structure (e.g. non-array `authors`/`tags`, missing fields) and the error is not contained per-repo.

**Expected behavior**
Repos with missing/invalid manifests should be skipped individually (they already are for 404s via the worker), and the card-building/render pipeline should be wrapped per-repo so one bad manifest can never crash the whole grid — especially since the topic search currently surfaces many dead repos.

**Impact**
Every extension listed on pages 2-4 (i.e. most new/star-less repos, including newly published ones) is impossible to reach: users crash the grid before the page loads. New extensions effectively cannot be installed through the Marketplace UI at all.

rxri:
Duplicate of https://github.com/spicetify/marketplace/issues/1203, also was the AI needed to write this issue?

Issue #1203, opened 2026-07-29
Crash during search: TypeError: Cannot read properties of undefined (reading 'toLowerCase')


GamerNation12:
### 🔍 Is there already an issue for your problem?

- [x] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version: 1.2.94.583.g60394bd5
- Spicetify version: v2.44.0
- OS: Windows 11
```

### 📝 Description

**What happens:**
When searching the marketplace and clicking "Load More", the client crashes with a fatal React error: `TypeError: Cannot read properties of undefined (reading 'toLowerCase')`. This happens when the search filter attempts to parse a repository item that is missing standard string fields (like `name`, `author`, or `tags`).

**Steps to reproduce:**
1. Open Spicetify Marketplace.
2. Search for an item.
3. Click "Load More" to fetch additional items.
4. If any fetched item has malformed or missing fields, the UI crashes.

**What I expected to happen:**
The marketplace should silently ignore or skip items with malformed data instead of crashing the entire client. 

**Proposed Fix / Extra Details:**
The search filter logic lacks type checking before calling `.toLowerCase()`. Adding optional chaining and `typeof` checks resolves the crash completely. 

Example patch for the filtering logic:
`!t || (typeof e === "string" && e.toLowerCase().includes(t)) || (typeof a === "string" && a.toLowerCase().includes(t)) || r?.some((e) => typeof e?.name === "string" && e.name.toLowerCase().includes(t)) || [...(n ?? [])].some((e) => typeof e === "string" && e.toLowerCase().includes(t))`

PR #1221, opened 2026-08-24
chore: release 1.0.10

ImNoammm:
## Summary

- bump Marketplace from 1.0.9 to 1.0.10
- trigger the existing auto-release workflow for fixes already merged since v1.0.9

## Why

The latest release predates the manifest validation merged in #1209. Release users can therefore still hit the Load More/search crash reported in #1203 and #1207 even though it is fixed on `main`.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm lint:ci`
- `pnpm build:local`
- live-tested the current `dist` containing #1209 on Spicetify 2.44.0: an active search remained stable while loading the affected extension pages

## Summary by CodeRabbit

* **Chores**
  * Updated the application package to version 1.0.10.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "marketplace",
-  "version": "1.0.9",
+  "version": "1.0.10",
   "private": true,
   "homepage": "https://github.com/spicetify/marketplace",
   "bugs": {


rxri:
There won't be marketplace update for now.

